PR #1, opened 2020-11-04
Switch deploy to docker-hub

luke-biel:


document-storage/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [[bin]]
-name = "document_storage"
+name = "sleigh-db"
```
frondeus, line 11: I guess we should change directory folder as well, right?

luke-biel: Yeah, I stopped mid-changes, consider this a draft (I couldn't mark it as a draft because tests don't run on drafts).
Only yaml structure is kinda final.
I'm waiting for decision on naming of blob, then I'll make adjustments.

USSAMOH2 (approved):

Kononnable (commented):
Are github actions protected from stealing secrets or is there something we have to do to achieve that?
What I mean is if someone can change github action definition in PR and this(changed) action will be executed with secrets. Some CI/CD systems drops secrets on actions started by not-trusted user and I want to be sure that's the case here.

Otherwise good to go.

luke-biel:
We don't run actions on external PR's and secrets are hidden if they were to show in the log.

Only scenario where a secret leaks is if anybody from within our organization would text-transform a secret and forced it to be printed out. Which nobody can do by an `accident`.

PR #4, opened 2020-11-05
Add payload to reported errors

### LINKED ISSUE #2: Add `payload` to report message JSON body
Changes needed in command_service are:
- Add message payload to JSON body (some clients require that we return original message if error happens)
- Make reporting functionality optional (we don't want to initialize reporting module by default)
- Make payload only available in VERBOSE reporting mode (default error reporting must not contain payload)

luke-biel:
Resolves #2


luke-biel:
PR is split into 2 commits - first is refactoring, second is adding `payload`

command-service/src/report/mod.rs
```diff
@@ -1,62 +1,39 @@
...
 
 mod config;
 mod error;
+mod verbose_report_service;
 
-pub struct ReportService {
-    producer: FutureProducer,
-    topic: String,
+pub enum ReportService {
+    Verbose(VerboseReportServiceConfig),
```
Kononnable, line 11: Not sure if verbose is a good name - right now it's the only option other then no reporting. Anyway it's just an internal name so no problems here.

Kononnable (commented):
Q: How much of this will be applicable after merging #5?

Please limit force pushes to minimum when PR is waiting for reviews - github errors with 'out of date diff' then

command-service/src/report/mod.rs
luke-biel: I renamed it to `Full`. Can't really think of any valid name.

As for #5, my changes remove channels from this module, which is exclusive to your changes, however I assume some rebasing will be necessary.

luke-biel:
> Please limit force pushes to minimum when PR is waiting for reviews - github errors with 'out of date diff' in such cases

I assumed nobody looked at it yet :)

esavier (commented):
looks okay, but since its kinda big commit, someone else also needs to approve it

luke-biel:
@USSAMOH2 Can you please review code? I guess nobody else can push this forward.

command-service/src/report/mod.rs
```diff
@@ -1,48 +1,39 @@
...
+            ReportService::Full(config) => Box::new(FullReportServiceInstance {
+                producer: config.producer.clone(),
+                topic: config.topic.clone(),
+                output_plugin: config.output_plugin.clone(),
+                msg: msg.clone(),
+            }),
+            ReportService::Disabled => Box::new(()),
+        }
+    }
+}
```
USSAMOH2, line 39: This has code smell written all over it. Please avoid defaulting to dynamic dispatch when possible, and either reimplement this as an enum or at least remove the need for this `instantiate` method to be called.

command-service/src/report/mod.rs
```diff
@@ -1,48 +1,39 @@
...
-            description.replace('"', r#"\""#),
-            object_id,
-        );
-
-        self.producer
-            .publish_message(&self.topic, "command_service.status", payload.into())
-            .await
-            .map_err(Error::FailedToReport)?;
+#[async_trait::async_trait]
+pub trait ReportServiceInstance: Send + Sync + 'static {
```
USSAMOH2, line 16: I assume you did this as a way to not have to rename the existing struct `ReportService`, but the trait should be a verb and the struct should be a noun, not the other way around. Please rename this trait to `ReportService` or `Reporter`, and then rename `ReportService` to something else.

command-service/src/report/full_report_service.rs
```diff
@@ -0,0 +1,57 @@
...
+
+const APPLICATION_NAME: &str = "Command Service";
+
+pub struct FullReportServiceConfig {
+    pub producer: CommonPublisher,
+    pub topic: Arc<String>,
+    pub output_plugin: Arc<String>,
+}
+
+pub struct FullReportServiceInstance {
```
USSAMOH2, line 15: Instance refers to a literal, singular value, not a type. Please remove the `Instance` suffix.

command-service/src/report/full_report_service.rs
```diff
@@ -0,0 +1,57 @@
...
+            topic: Arc::new(topic),
+            output_plugin: Arc::new(output_plugin),
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl ReportServiceInstance for FullReportServiceInstance {
+    async fn report(&mut self, description: &str) -> Result<(), Error> {
+        let payload = json!({
```
USSAMOH2, line 37: Though it will technically make the code longer, we should avoid allocating a dynamic JSON object and then rendering it to a string every time (I'd be surprised if that was optimized away, as good as LLVM is). This can be avoided by deriving Serialize on a struct with a lifetime like so:

```rust
#[derive(Serialize)]
pub struct ReportMessage<'a> {
    pub application: &'static str,
    ...
}
```

command-service/src/report/full_report_service.rs
```diff
@@ -0,0 +1,57 @@
...
+        self.producer
+            .publish_message(
+                self.topic.as_str(),
+                "command_service.status",
+                payload.into_bytes(),
+            )
+            .await
+            .map_err(Error::FailedToReport)?;
+
+        Ok(())
```
USSAMOH2, line 55: I know it's a common pattern to have an `Ok(())` at the end of the function, but you could just remove the `?;` and not even need the extra 2 lines.

command-service/src/report/full_report_service.rs
```diff
@@ -0,0 +1,57 @@
+use crate::communication::GenericMessage;
+use crate::report::{Error, ReportServiceInstance};
+use serde_json::json;
+use std::sync::Arc;
+use utils::messaging_system::publisher::CommonPublisher;
+
+const APPLICATION_NAME: &str = "Command Service";
+
+pub struct FullReportServiceConfig {
```
USSAMOH2, line 9: Should this derive Clone?

command-service/src/output/victoria_metrics/mod.rs
```diff
@@ -50,29 +46,17 @@ impl VictoriaMetricsOutputPlugin {
...
 
-#[async_trait]
+#[async_trait::async_trait]
 impl OutputPlugin for VictoriaMetricsOutputPlugin {
-    async fn handle_message(
-        &self,
-        recv_msg_bundle: ReceivedMessageBundle,
-    ) -> Result<(), OutputError> {
+    async fn handle_message(&self, msg: GenericMessage) -> Result<Resolution, OutputError> {
         let mut url = self.url.clone();
```
USSAMOH2, line 52: You shouldn't need to `.clone()` these anymore now that we aren't spawning a future to handle them.

command-service/src/output/victoria_metrics/mod.rs
```diff
@@ -50,29 +46,17 @@ impl VictoriaMetricsOutputPlugin {
...
 
-            let resolution = process_message(url, client, msg).await;
+        url.set_query(Some(&format!("db={}", msg.schema_id)));
 
-            if status_sender.send(resolution).is_err() {
-                error!("Failed to send status to report service");
-                status_endpoints::mark_as_unhealthy();
-            }
-        });
+        let resolution = process_message(url, client, msg).await;
```
USSAMOH2, line 57: Why not just wrap the resolution in an Ok?

command-service/src/output/mod.rs
```diff
@@ -27,66 +22,8 @@ pub enum OutputArgs {
...
-
-#[async_trait]
-pub trait OutputPlugin {
-    async fn handle_message(
-        &self,
-        recv_msg_bundle: ReceivedMessageBundle,
-    ) -> Result<(), OutputError>;
+#[async_trait::async_trait]
+pub trait OutputPlugin: Send + Sync + 'static {
+    async fn handle_message(&self, msg: GenericMessage) -> Result<Resolution, OutputError>;
```
USSAMOH2, line 27: In the scope of this PR, there is no way for any of the implementors of this method to return an error, only a bad resolution. This should be changed (along with all of its implementors) to either return errors when appropriate or to simply return a resolution.

command-service/src/output/druid/mod.rs
```diff
@@ -90,3 +56,19 @@ impl DruidOutputPlugin {
         }
     }
 }
+
+#[async_trait::async_trait]
+impl OutputPlugin for DruidOutputPlugin {
+    async fn handle_message(&self, msg: GenericMessage) -> Result<Resolution, OutputError> {
+        let producer = self.producer.clone();
```
USSAMOH2, line 63: Do these still need to be cloned with the lack of future spawning?

command-service/src/main.rs
```diff
@@ -1,41 +1,85 @@
...
+            start_services(
+                args.input_config,
+                args.report_config,
+                VictoriaMetricsOutputPlugin::new(victoria_metrics_config)
+                    .expect("Initializing Victoria Metrics output interface"),
+            )
+            .await
+        }
+    }
+    .expect("Running services")
```
USSAMOH2, line 62: Why not just return an error from `main`?

USSAMOH2 (changes requested):

command-service/src/output/psql/mod.rs
```diff
@@ -48,11 +44,7 @@ impl PostgresOutputPlugin {
         let connection = pool.get().await.unwrap();
         let payload: Value = match serde_json::from_slice(&msg.payload) {
```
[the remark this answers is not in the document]
luke-biel: Can't do, method returns Resolution which is not Try (and I'm not sure I want to implement Try for it)

command-service/src/report/config.rs
```diff
@@ -3,7 +3,7 @@ use structopt::StructOpt;
 #[derive(Clone, Debug, StructOpt)]
 pub struct ReportServiceConfig {
     #[structopt(name = "report-broker", long = "report-broker", env = "REPORT_BROKER")]
-    pub broker: String,
+    pub broker: Option<String>,
```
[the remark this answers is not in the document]
luke-biel: If env is not set it evaluates to None.

```rust
use structopt::StructOpt;

#[derive(StructOpt, Debug)]
struct TestEnv {
    #[structopt(env = "HOME")]
    home: Option<String>,
    #[structopt(env = "NOHOME")]
    no_home: Option<String>,
}

fn main() {
    let test = TestEnv::from_args();
    
    println!("{:?}", test);
}
```

command-service/src/report/mod.rs
luke-biel: This is Null Object Pattern.
I had real issue to come up with a way to not allocate memory when report service is disabled, but to allow message to be passed if it's enabled. I would love to discuss this if you see different approach, but it's as clean as I could do alone.

command-service/src/main.rs
luke-biel: I guess it wasn't there when I started this feature. We need to standarize this, we have some apps unwrap and some `?`

command-service/src/report/mod.rs
luke-biel: Naming things is difficult... Well, I changed them, you can verify if they are ok.

command-service/src/report/full_report_service.rs
luke-biel: Changed

command-service/src/report/full_report_service.rs
luke-biel: Good idea, done that.

command-service/src/report/full_report_service.rs
luke-biel: I don't see reason why except for `standard` that all struct should derive Debug and Clone. Well, I've added Clone.

command-service/src/output/victoria_metrics/mod.rs
luke-biel: Url has to be cloned sadly. But I checked other variables that I clone, and removed unnecessary.

command-service/src/output/victoria_metrics/mod.rs
luke-biel: Good point

command-service/src/output/mod.rs
luke-biel: Agree, good find. Changed Result to Resolution

command-service/src/output/druid/mod.rs
luke-biel: It doesn't have to.

command-service/src/main.rs
luke-biel: Changed to anyhow

esavier (dismissed):

command-service/src/report/full_report_service.rs
luke-biel: Changed

command-service/src/output/sleigh/mod.rs
```diff
@@ -55,38 +50,23 @@ impl SleighOutputPlugin {
...
-                        object_id: recv_msg_bundle.msg.object_id,
-                    }
-                }
-            };
-
-            if recv_msg_bundle.status_sender.send(resolution).is_err() {
-                error!("Failed to send status to report service");
-                status_endpoints::mark_as_unhealthy();
+    async fn handle_message(&self, msg: GenericMessage) -> Resolution {
+        let resolution = match self.pool.get().await {
```
USSAMOH2, line 61: You explicitly handle the error when getting a pool connection here, but `unwrap` when getting it elsewhere. You should pick one.

USSAMOH2, line 61: No need to wrap this in another scope

USSAMOH2 (changes requested):

command-service/src/output/sleigh/mod.rs
luke-biel: I totally missed that leftover unwrap(). Good find. I've removed it.

luke-biel: I think I did what you meant by that 🤔

esavier (approved):

USSAMOH2 (approved):

command-service/src/input/service.rs
```diff
@@ -1,33 +1,33 @@
...
+use tokio::pin;
+use utils::message_types::CommandServiceInsertMessage;
+use utils::messaging_system::consumer::CommonConsumer;
+use utils::messaging_system::message::CommunicationMessage;
+use utils::messaging_system::CommunicationResult;
+use utils::metrics::counter;
+use utils::task_limiter::TaskLimiter;
 
-pub struct KafkaInput {
+pub struct KafkaInput<P: OutputPlugin> {
```
Kononnable, line 15: Wasn't the whole idea of plugins architecture to keep input and output as separate entities?

command-service/src/input/service.rs
```diff
@@ -77,6 +79,7 @@ impl KafkaInput {
                 .run(async move || {
                     if let Err(err) = Self::handle_message(router, message).await {
                         error!("Failed to handle message: {}", err);
+                        process::abort();
```
Kononnable, line 82: Why is this critical? Failing to save a specific message doesn't really mean that whole service is in an unknown state.
There is a question of kafka changing offsets instead of acknowledging each message on its own but as I recall we've got it 'handled' by `enable.auto.commit`

command-service/src/report/full_report_sender.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(Clone)]
+pub struct FullReportSenderBase {
+    pub producer: CommonPublisher,
+    pub topic: Arc<String>,
+    pub output_plugin: Arc<String>,
+}
+
+#[derive(Clone)]
+pub struct FullReportSender {
```
Kononnable, line 19: I think this should rather contain `FullReportSenderBase` instead of repeating those fields again and again(fields are small so memory footprint isn't a big issue probably).

command-service/src/report/full_report_sender.rs
```diff
@@ -0,0 +1,67 @@
...
+            output_plugin: self.output_plugin.as_str(),
+            description,
+            object_id: self.msg.object_id,
+            payload: String::from_utf8_lossy(&self.msg.payload),
+        };
+
+        self.producer
+            .publish_message(
+                self.topic.as_str(),
+                "command_service.status",
```
Kononnable, line 61: Can be extracted as const.

PR #5, opened 2020-11-06
abstraction over messaging queues

Kononnable:
- abstraction over communication method
   - allows easy change from kafka to rabbit and vice versa - only consumer/publisher initialization is message system specific
   - doesn't support runtime/compile flag selection of which messaging system we want to use
- some code cleanup(e.g. unused feature flags, unused code)

TODO(outside of this PR scope):
- a way to change messaging system without making code changes - feature flags, runtime check or something else
- there might be some obsolete method/variable names

I'm not exactly sure how to handle errors here. Currently I'm converting each error to common error type - alternatively we can use same approach as in command service(different error type for each error) or deal with that some other way.

### DIFF
--- a/command-service/src/input/service.rs
+++ b/command-service/src/input/service.rs
@@ -66,7 +66,7 @@ impl KafkaInput {
     }
 
     pub async fn listen(self) -> Result<(), Error> {
-        let consumer = Box::leak(Box::new(self.consumer));
+        let consumer = self.consumer.leak();
         let message_stream = consumer.consume().await;
         pin!(message_stream);
 
--- a/command-service/src/input/service.rs
+++ b/command-service/src/input/service.rs
@@ -1,42 +1,35 @@
-use std::sync::Arc;
-
 use futures_util::stream::StreamExt;
-use log::{error, log_enabled, trace, Level};
-use rdkafka::config::RDKafkaLogLevel;
-use rdkafka::consumer::{Consumer, DefaultConsumerContext, StreamConsumer};
-use rdkafka::error::KafkaError;
-use rdkafka::message::{Headers, OwnedMessage};
-use rdkafka::{ClientConfig, Message};
-use uuid::Uuid;
+use log::error;
+use tokio::pin;
 
 use crate::communication::{GenericMessage, MessageRouter};
 use crate::input::{Error, KafkaInputConfig};
-use utils::{metrics::counter, task_limiter::TaskLimiter};
+use utils::{
+    message_types::CommandServiceInsertMessage,
+    messaging_system::{
+        consumer::CommonConsumer, message::CommunicationMessage, CommunicationResult,
+    },
+    metrics::counter,
+    task_limiter::TaskLimiter,
+};
 
 pub struct KafkaInput {
-    consumer: Arc<StreamConsumer<DefaultConsumerContext>>,
+    consumer: CommonConsumer,
     message_router: MessageRouter,
     task_limiter: TaskLimiter,
 }
 
 impl KafkaInput {
-    pub fn new(config: KafkaInputConfig, message_router: MessageRouter) -> Result<Self, Error> {
-        let consumer: StreamConsumer<DefaultConsumerContext> = ClientConfig::default()
-            .set("group.id", &config.group_id)
-            .set("bootstrap.servers", &config.brokers)
-            .set("enable.partition.eof", "false")
-            .set("session.timeout.ms", "6000")
-            .set("enable.auto.commit", "true")
-            .set_log_level(RDKafkaLogLevel::Debug)
-            .create_with_context(DefaultConsumerContext)
-            .map_err(Error::ConsumerCreationFailed)?;
-
-        consumer
-            .subscribe(&[&config.topic])
-            .map_err(Error::FailedToSubscribe)?;
-
+    pub async fn new(
+        config: KafkaInputConfig,
+        message_router: MessageRouter,
+    ) -> Result<Self, Error> {
+        let consumer =
+            CommonConsumer::new_kafka(&config.group_id, &config.brokers, &[&config.topic])
+                .await
+                .map_err(Error::ConsumerCreationFailed)?;
         Ok(Self {
-            consumer: Arc::new(consumer),
+            consumer,
             message_router,
             task_limiter: TaskLimiter::new(config.task_limit),
         })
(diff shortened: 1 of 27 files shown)


command-service/src/input/service.rs
```diff
@@ -59,65 +52,26 @@ impl KafkaInput {
...
-                .map_err(|_err| Error::InvalidSchemaIdHeader)
-        } else {
-            Err(Error::MissingSchemaIdHeader)
-        }
-    }
-
-    pub async fn listen(&self) -> Result<(), Error> {
-        let mut message_stream = self.consumer.start();
+    pub async fn listen(self) -> Result<(), Error> {
+        let consumer = Box::leak(Box::new(self.consumer));
```
Kononnable, line 69: Not the nicest thing to do, but consumer needs to have static lifetime(or be passed everywhere) and it's static according to application logic anyway.

data-router/src/main.rs
```diff
@@ -37,115 +30,94 @@ struct Config {
...
+async fn handle_message(
+    message: Box<dyn CommunicationMessage>,
+    cache: Arc<Mutex<LruCache<Uuid, String>>>,
+    producer: Arc<CommonPublisher>,
+    config: Arc<Config>,
+) {
+    let result: anyhow::Result<()> = async {
+        let event: DataRouterInputData =
+            serde_json::from_str(message.payload()?).context("Payload deserialization failed")?;
+        let since_the_epoch = SystemTime::now()
```
luke-biel, line 79: We probably should allow this data to have timestamp already, instead of always inserting it.

data-router/src/main.rs
```diff
@@ -37,115 +30,94 @@ struct Config {
...
-
-    metrics::serve();
-    message_handler.await?;
-
+                    config.clone(),
+                ));
+            }
+            Err(error) => {
+                error!("Error fetching data from message queue {:?}", error);
+                // no error handling necessary - message won't be acked - it was never delivered properly
```
luke-biel, line 63: This is fine for now, but we definitely need to report these incidents somewhere.
- have some log scrapper
- have counter go up in metrics
- send info to report service

command-service/src/input/service.rs
luke-biel: I would see this in one, well documented place, because someone may miss this leak in the future otherwise.
Options are -
- move creating consumer to main, and there leak it (we will have such behaviour only on top level, it's easier to debug, nobody accidentaly puts this in a loop)
- create method within consumer (or utils in general), eg.`elevate_lifetime_to_static`, that would be explicit in what it does, but is done once in whole codebase - then we can replace it easier
- try and put that in lazy_static or Once at top of this file

I'm leaning towards 2nd option, it's fastest and since you leak this consumer more than once (in data router too), it would merge it all

utils/src/messaging_system/message.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#[async_trait]
+pub trait CommunicationMessage: Send + Sync {
+    fn payload(&self) -> CommunicationResult<&str>;
+    fn key(&self) -> CommunicationResult<&str>;
+    async fn ack(&self) -> CommunicationResult<()>;
+}
+
+pub struct KafkaCommunicationMessage<'a> {
+    pub(super) message: BorrowedMessage<'a>,
```
luke-biel, line 22: I'd say we just use pub and limit visibility in parent directory.

luke-biel (changes requested):
Only `leak` comment is kinda blocking. Otherwise you may merge it in my opinion.

utils/src/messaging_system/consumer.rs
```diff
@@ -0,0 +1,102 @@
...
+    message::RabbitCommunicationMessage, CommunicationResult,
+};
+
+pub enum CommonConsumer {
+    Kafka {
+        consumer: Arc<StreamConsumer<DefaultConsumerContext>>,
+    },
+    RabbitMq {
+        _connection: Box<Connection>,
+        _channel: Box<Channel>,
```
frondeus, line 23: Why do we wanna store `_connection` and `_channel`? As far as I remember since lapin 1.3 `Channel` and `Connection` are not closed automatically if there is active consumer.

data-router/src/main.rs
```diff
@@ -37,115 +30,94 @@ struct Config {
...
 async fn main() -> anyhow::Result<()> {
     env_logger::init();
     let config = Arc::new(envy::from_env::<Config>().context("Env vars not set correctly")?);
+    metrics::serve();
 
-    let (_conn, _channel, mut consumer) =
-        create_rabbitmq_consumer(&config.input_addr, &config.input_queue).await?;
-    let producer = Arc::new(build_kafka_producer(&config.kafka_brokers));
+    let consumer =
+        CommonConsumer::new_rabbit(&config.input_addr, "CDL_DATA_ROUTER", &config.input_queue)
```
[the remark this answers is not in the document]
Kononnable: Moved to const. There is no need for configuration variable since it's 'internal' data - no one should be interested in it unless he/she is debugging.

utils/src/messaging_system/consumer.rs
Kononnable: Yep, legacy reasons. Removed.

data-router/src/main.rs
Kononnable: Yeah, we definitively need to change that in the future. The question 'how' is kind of problematic. Requiring value from external systems(users) can be problematic since we have no idea if their clocks are synchronized(even if there are there can be some timing problems in specific scenarios). We have no idea what would happen if someone enters timestamp from the past(we have not tested that scenario).

For now I'll leave it be but that need to be changed in the future.

luke-biel (approved):

data-router/src/main.rs
Kononnable: It's 'internal' error so I wouldn't go for report service(users doesn't really care about error which can be automatically fixed).
Log scrapping is already part of epiphany, if someone decides to deploy cdl on its own they can use `kubectl logs` if they don't use custom logging scrapper.

As for metrics - I think this is a good idea, but currently we have no other similar metrics(debug/performance monitoring centered) - only user centered(throughput).

utils/src/messaging_system/message.rs
Kononnable: You mean removing `pub` from `pub mod message;` and just reexport the trait from this module? 
My intention here was to allow types like `KafkaCommunicationMessage` to be accessed from outside, but forbid creating new values of such types in case there are some method on such types which aren't part of the trait - we still don't know how we're gonna decide which messaging we're using(in runtime or by feature flags or similar construct).

luke-biel:
@Kononnable before you are gone? Can we merge it? Assuming nobody speaks up that something is wrong.

PR #6, opened 2020-11-06
update documentation

rlawrence25:


README.md
```diff
@@ -60,13 +50,12 @@ examples/deploy | sample deployment guide for docker
...
 
 
 [rust]: https://www.rust-lang.org
 [sled]: https://github.com/spacejam/sled
 [kafka]: https://kafka.apache.org/
 [rmq]: https://www.rabbitmq.com/
 [schema registry]: ./schema-registry/
-[confluence wiki]: https://pgga-es.atlassian.net/wiki/spaces/EPI/pages/602938021/Common+Data+Layer+concept
-[Getting Started Wiki]: https://github.com/pgga-es/epiphany-cdl-rust/wiki/Getting-Started
+[Getting Started]: https://github.com/epiphany-platform/CommonDataLayer/blob/develop/docs/K8s-Local-Deployment.md
```
Kononnable, line 61:
```suggestion
[Getting Started]: ./docs/K8s-Local-Deployment.md
```
It's better to use relative paths otherwise it may change git branch person is looking on.
Not sure if we can use how to setup deployment as a Getting Started(no direct info how to work with the project - like create schema, inject data in this specific format etc.)  but I think it's good enough for now(at least until we decide to make documentation for much more stuff).

rlawrence25: Whoops. Nice catch. I actually created a dedicated "Getting Started" markdown file which is pretty bare and just serves as a directory to other resources and files. So the link is wrong here. As for working with Schema, should I link the CLI's documentation since some of that information lives there?

Kononnable: I think linking CLI docs for now is fine - it will at least give someone a starting point. They would have to do most of the heavy lifting themselves, but at least they will know where to start.
We'll need to probably write some 'walk through' example one day before anyone can take an interest in cdl, but if it is better to do it now, or in a month I have no idea.

luke-biel (approved):

Kononnable (approved):

PR #33, opened 2020-11-19
Streamline dockerfile

Tesel:
1. Make dockerfile smaller to make it more maintainable and uses less commands so it's faster
2. Get rid of musl in place of dynamically linked glibc, 
It improves build time by ~30s and might improve performance: https://andygrove.io/2020/05/why-musl-extremely-slow/


USSAMOH2 (approved):

Kononnable:
I'm going to revert this.
 - base branch is not develop, is some other feature branch which is WIP
 - > and uses less commands so it's faster

   Not really faster because of fewer commands. Those commands are necessary for docker to be able to cache produced results.  With 'old' build approach after first (full) build if you change just one file it will only rebuild necessary crates. If you use `COPY . ./` it will invalidate docker cache(because whole layer is different) which means downloading each dependency from crates.io, compiling them all etc. 
With old approach full rebuild is done only when updating dependencies - it's not  targeted for shorter time of full build, caching is the benefit here.

PR #34, opened 2020-11-19
feat(data-router): ingest messages from REST as well in the Data Router

USSAMOH2:


luke-biel (approved):
Two notes:
I'm not sure it was necessary to add REST for feature to function (but with rest it will be more consistent)
I don't know if it's good that this PR has two authors

Otherwise LGTM

esavier (changes requested):
DOCUMENTATION!!!!11

Kononnable:
Tbh. I don't like it. 
- there was no feature request from future clients
- it's always on - kafka, rabbitmq can be used interchangeably, same will go for protocol used in service mesh(as far as I understood), why rest is special?

PR #35, opened 2020-11-23
feat(schema-registry): add methods for exporting and importing all data from DB to json

frondeus:
TODO:
* [ ] Documentation
* [ ] Manual tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,4 @@
 registry_db/
 storage_db/
 target/
-.vscode/
-.idea
 venv
-.envrc
-shell.nix
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ target/
 .vscode/
 .idea
 venv
+.envrc
+shell.nix
--- a/schema-registry/src/error.rs
+++ b/schema-registry/src/error.rs
@@ -54,13 +54,10 @@ impl From<jsonschema::CompilationError> for RegistryError {
 pub enum MalformedError {
     #[error("Schema in database was malformed with id {0}")]
     MalformedSchema(Uuid),
-
     #[error("Schema definition in database was malformed with id {0}")]
     MalformedDefinition(Uuid),
-
     #[error("Schema version in database was malformed under schema with id {0}")]
     MalformedSchemaVersion(Uuid),
-
     #[error("View in database was malformed for view with id {0}")]
     MalformedView(Uuid),
 }
--- a/schema-registry/src/error.rs
+++ b/schema-registry/src/error.rs
@@ -54,8 +54,13 @@ impl From<jsonschema::CompilationError> for RegistryError {
 pub enum MalformedError {
     #[error("Schema in database was malformed with id {0}")]
     MalformedSchema(Uuid),
+
+    #[error("Schema definition in database was malformed with id {0}")]
+    MalformedDefinition(Uuid),
+
     #[error("Schema version in database was malformed under schema with id {0}")]
     MalformedSchemaVersion(Uuid),
+
     #[error("View in database was malformed for view with id {0}")]
     MalformedView(Uuid),
 }
--- a/schema-registry/src/rpc.rs
+++ b/schema-registry/src/rpc.rs
@@ -62,13 +62,11 @@ impl SchemaRegistryImpl {
     }
 
     pub fn export_all(&self) -> anyhow::Result<DbExport> {
-        let result = self.db.export_all()?;
-        Ok(result)
+        self.db.export_all().context("Failed to export the entire database")
     }
 
     pub fn import_all(&self, imported: DbExport) -> anyhow::Result<()> {
-        self.db.import_all(imported)?;
-        Ok(())
+        self.db.import_all(imported).context("failed to import database")
     }
 }
 
--- a/schema-registry/src/rpc.rs
+++ b/schema-registry/src/rpc.rs
@@ -65,6 +65,11 @@ impl SchemaRegistryImpl {
         let result = self.db.export_all()?;
         Ok(result)
     }
+
+    pub fn import_all(&self, imported: DbExport) -> anyhow::Result<()> {
+        self.db.import_all(imported)?;
+        Ok(())
+    }
 }
 
 #[tonic::async_trait]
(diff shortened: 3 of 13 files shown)


.gitignore
```diff
@@ -1,6 +1,4 @@
 registry_db/
 storage_db/
 target/
-.vscode/
```
USSAMOH2, line 4: Why would we want these things committed?

schema-registry/src/error.rs
```diff
@@ -54,8 +54,13 @@ impl From<jsonschema::CompilationError> for RegistryError {
 pub enum MalformedError {
     #[error("Schema in database was malformed with id {0}")]
     MalformedSchema(Uuid),
+
```
USSAMOH2, line 57: What's with the extra space? I personally think we don't need it, but if you would prefer to have it, we should be consistent across all of our error enums.

schema-registry/src/rpc.rs
```diff
@@ -59,6 +60,16 @@ impl SchemaRegistryImpl {
             .unwrap_or_else(abort_on_poison)
             .replicate_message(event)
     }
+
+    pub fn export_all(&self) -> anyhow::Result<DbExport> {
+        let result = self.db.export_all()?;
+        Ok(result)
```
USSAMOH2, line 66:
```suggestion
        self.db.export_all().context("Failed to export the entire database")
```

schema-registry/src/rpc.rs
```diff
@@ -59,6 +60,16 @@ impl SchemaRegistryImpl {
...
+    }
+
+    pub fn import_all(&self, imported: DbExport) -> anyhow::Result<()> {
+        self.db.import_all(imported)?;
+        Ok(())
```
USSAMOH2, line 71:
```suggestion
        self.db.import_all(imported).context("failed to import database")
```

schema-registry/src/types/storage/edges.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    fn from_properties(properties: EdgeProperties) -> Option<Self> {
+        let schema_id = properties.edge.key.outbound_id;
+        let view_id = properties.edge.key.inbound_id;
+        Some(Self { schema_id, view_id })
+    }
+
+    fn db_type() -> Type {
+        SCHEMA_VIEW_EDGE_TYPE.clone()
```
USSAMOH2, line 36: Do you need the `lazy_static!` definition if you only really need them here?

schema-registry/src/types/storage/edges.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct SchemaDefinition {
+    pub schema_id: Uuid,
+    pub def_id: Uuid,
```
USSAMOH2, line 43: I know some people prefer brevity, but I would say that the elision here from `definition` to `def` makes the code less readable. How strongly do you prefer the shorter version?

schema-registry/src/types.rs
```diff
@@ -79,3 +80,35 @@ impl VersionedUuid {
...
+#[derive(Default, Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct DbExport {
+    pub schemas: HashMap<Uuid, vertices::Schema>,
+    pub definitions: HashMap<Uuid, vertices::Definition>,
+    pub views: HashMap<Uuid, vertices::View>,
+    pub def_edges: Vec<edges::SchemaDefinition>,
+    pub view_edges: Vec<edges::SchemaView>,
+}
+
+fn get_vertex_property_or<T: DeserializeOwned>(
```
USSAMOH2, line 94:
```suggestion
fn extract_vertex_property<T: DeserializeOwned>(
```
To imply that we are removing the desired property.

schema-registry/src/types.rs
```diff
@@ -79,3 +80,35 @@ impl VersionedUuid {
...
+    name: &'static str,
+) -> Option<T> {
+    properties
+        .props
+        .drain_filter(|prop| prop.name == name)
+        .next()
+        .and_then(|prop| serde_json::from_value(prop.value).ok())
+}
+
+fn get_edge_property_or<T: DeserializeOwned>(
```
USSAMOH2, line 105:
```suggestion
fn extract_edge_property<T: DeserializeOwned>(
```

USSAMOH2 (changes requested):

schema-registry/src/types/storage/edges.rs
```diff
@@ -0,0 +1,72 @@
+use super::*;
```
luke-biel, line 1: Not a fan

.gitignore
frondeus: We don't!
But as @luke-biel pointed, its committer responsibility to have these folders in his own personal global `.gitignore`.

schema-registry/src/error.rs
frondeus: Well for me its easier to read

schema-registry/src/types/storage/edges.rs
frondeus: As far as I understand `Type::new()` has some validation which we unwrap in lazy_static.
I could use constructor here and forget about it but then we would validate on every callsite.
I we don't care about it then yeah, we can remove lazy_static.

luke-biel: We can just create Type with unchecked, we are guaranteed these are correct.

luke-biel:
@frondeus docs aren't as important as other parts of this commit, so you can add them in separate PR.

schema-registry/src/types/storage/edges.rs
frondeus: But then we `.clone()` - there is allocation anyway right?

frondeus:
> @frondeus docs aren't as important as other parts of this commit, so you can add them in separate PR.

I know, that's why i focus on other parts than docs.

schema-registry/src/types/storage/edges.rs
USSAMOH2: I'm fine with leaving it as is, because the `unchecked` alternative uses unsafe.

USSAMOH2 (approved):

PR #42, opened 2020-11-26
fix(data-router): Change input consumer from Rabbit to Kafka

frondeus:


luke-biel (commented):
Fix helm and docker-compose along

frondeus:
> Fix helm and docker-compose along

Ah! right!

frondeus:
TODO:
- [ ] Manual tests

helm/helm-infrastructure/Chart.yaml
```diff
@@ -23,9 +23,6 @@ version: 0.0.5
 appVersion: 0.0.5
 
 dependencies:
-- name: rabbitmq
```
Kononnable, line 26: If we remove this we should also regenerate `Chart.lock` file. 

tbh. I'm ok with keeping it for now - if we're gonna remove rabbitmq entirely we could also remove code referring to it. Anyway, this can be handled later on.

frondeus: Okay, I agree. Restored file

Kononnable (approved):

esavier (approved):

PR #49, opened 2020-11-27
chore: summarized fixes from dev

luke-biel:


esavier (approved):
approved, but about pushing common version as tag to all crates - we should pick it up later

Kononnable (approved):

esavier (approved):

Issue #15, opened 2020-11-17
Query Router - discern Timeseries (range query). and Document/Binary (access via UUID)


luke-biel:
This requires change in schema-registry, so we can distinguish what type of repository are we querying.

We must add new type of node to indra-db that will connect to schema_id and would describe TYPE of repository (document or timeseries).
We must add gRPC endpoint that allows us to query for TYPE given schema_id.

Acceptance criteria:
* [x] SchemaRegistry can contain ONE repo *TYPE* per schema_id
* [x] We can add *TYPE* to existing schema over gRPC
* [x] We can query *TYPE* over gRPC
* [x] We can add, remove and update *TYPE* via `cdl-cli`
* [x] We have to integration test `query-service-ts` in python

PR #56, opened 2020-11-30
feat(schema-registry): store type of schema

damszew:
This change:
- extends `schema-registry` so it stores *TYPE* of schema per schema_id
- adds necessary gRPC methods to `schema-registry` to allow malipulations of schema *TYPE*
- adds subcommands to `cdl-cli` to use gRPC methods of `schema-registry`

This is only a part of #15 - `query-router` part will be done in a separate PR when `query-service-ts` is delivered.


frondeus (dismissed):

schema-registry/src/types.rs
```diff
@@ -19,6 +20,33 @@ pub struct NewSchema {
...
+}
+
+impl std::fmt::Display for SchemaType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let printable = match self {
+            SchemaType::DocumentStorage => "DocumentStorage",
+            SchemaType::Timeseries => "Timeseries",
+        };
+
+        write!(f, "{}", printable)
```
luke-biel, line 48:
```suggestion
        f.write_str(match self {
            SchemaType::DocumentStorage => "DocumentStorage",
            SchemaType::Timeseries => "Timeseries",
        })```

**Review by luke-biel — DISMISSED** [review]

File: schema-registry/src/types.rs (hunk 1 above)
Comment (damszew) [reply schema-registry/src/types.rs]: ok

File: cdl-cli/src/actions/schema.rs [hunk 2]
```diff
@@ -35,6 +35,7 @@ pub async fn add_schema(
     query_address: String,
     file: Option<PathBuf>,
     registry_addr: String,
+    schema_type: i32,
```
Kononnable, line 38: Can we use enum type here? This would force simple argument checking on cli tool side.

cdl-cli/src/actions/schema.rs
```diff
@@ -130,6 +132,22 @@ pub async fn set_schema_query_address(
     Ok(())
 }
 
+pub async fn set_schema_type(
+    schema_id: Uuid,
+    schema_type: i32,
```
Kononnable, line 137: As above

Kononnable (commented):
Apart from this one issue LGTM.

cdl-cli/src/actions/schema.rs
damszew: It could be done by implementing `FromStr` trait for enum but this enum would still need to be casted to `i32` to be send over gRPC

Kononnable (dismissed):

luke-biel (dismissed):

esavier (dismissed):

cdl-cli/src/actions/schema.rs
USSAMOH2: I still think that having the enum that is concerted before sending sets a smaller cardinality of the possible inputs for the CLI, and "self-documents" the options.

cdl-cli/src/actions/schema.rs
```diff
@@ -197,6 +215,19 @@ pub async fn get_schema_query_address(
...
+    let mut client = connect_to_registry(registry_addr).await?;
+    let response = client
+        .get_schema_type(Id {
+            id: schema_id.to_string(),
+        })
+        .await?;
+
+    println!("{}", response.into_inner().schema_type);
+
+    Ok(())
```
USSAMOH2, line 228: Shouldn't this return the schema type?

cdl-cli/src/args.rs
```diff
@@ -85,6 +92,9 @@ pub enum SchemaAction {
         /// the schema will be read from stdin.
         #[structopt(short, long, parse(from_os_str))]
         file: Option<PathBuf>,
+        /// The type of schema. Possible values: 0 (DocumentStorage), 1 (Timeseries).
+        #[structopt(short, long, default_value = "0")]
+        schema_type: i32,
```
USSAMOH2, line 97: As mentioned above, this exposes implementation details and is less clean than an enum, so please consider using an enum for the SchemaType.

cdl-cli/src/args.rs
```diff
@@ -131,6 +141,16 @@ pub enum SchemaAction {
         query_address: String,
     },
 
+    /// Update a schema's type in the registry.
+    SetSchemaType {
+        /// The id of the schema.
+        #[structopt(short, long)]
+        id: Uuid,
+        /// The new type of the schema. Possible values: 0 (DocumentStorage), 1 (Timeseries).
+        #[structopt(short, long)]
```
USSAMOH2, line 150: See above.

USSAMOH2 (changes requested):

USSAMOH2:
You should also make sure that backing up and restoring the schema registry saves the schema types as well.

cdl-cli/src/actions/schema.rs
damszew: Ok, in that case I'll use enum here

cdl-cli/src/actions/schema.rs
damszew: No because we are returning only if operation was successful or not. Schema type is shown to user via `println` (same as in other functions).

cdl-cli/src/args.rs
damszew: ok, I will use enum

damszew:
@USSAMOH2 There shouldn't be any problems with import/export as unit tests cover this functionality. Nevertheless I have manually tested it and it works as expected (After I've added creation of file before writing export data to it).

frondeus (approved):
LGTM

Kononnable (dismissed):

schema-registry/src/types.rs
```diff
@@ -19,6 +20,43 @@ pub struct NewSchema {
     pub definition: Value,
     pub kafka_topic: String,
     pub query_address: String,
+    pub schema_type: SchemaType,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub enum SchemaType {
```
USSAMOH2, line 27: Because you cast to an i32, you should maybe use repr(i32) or implement Into<i32> to ensure that the correct type is used.

USSAMOH2 (approved):
Other than the i32 thing, this looks great!

schema-registry/src/types.rs
damszew: All right, I will add `repr(i32)`

smores56 (approved):

luke-biel (approved):

PR #57, opened 2020-11-30
feat(query-service): Add RAW path to Postgres

MrDarthShoe:
Added RAW path to query-service: 
-Postgresql- manually tested
-Druid- not implemented
-Sled- not implemented


esavier (approved):
Good for me, but let's wait for others

esavier (dismissed):

query-service/proto/query.proto
```diff
@@ -17,3 +18,7 @@ message SchemaId {
 message ValueMap {
     map<string, bytes> values = 1;
 }
+
+message RawMsg {
+    string raw_msg = 1;
```
Kononnable, line 23: Raw message contains raw message?
Statement, query or something similar might be a better name.

query-service/src/psql.rs
```diff
@@ -123,4 +123,13 @@ impl Query for PsqlQuery {
             values: Self::collect_id_payload_rows(rows),
         }))
     }
+
+    async fn query_raw(&self, request: Request<RawMsg>) -> Result<Response<ValueMap>, Status> {
+        counter!("cdl.query-service.query_raw.psql", 1);
+        let rows = self.make_query(&request.into_inner().raw_msg, &[]).await?;
```
USSAMOH2, line 129: Since this expects the query to return the same format (not just arbitrary data), this should either document the expected output format somewhere, or use a different response type to allow for such arbitrary data.

query-service/proto/query.proto
MrDarthShoe: In terms of the same naming I see a convention here: like ObjectIds have object_ids, SchemaID has schema_id-> so RawMsg has raw_msg.
But I agree- I didn't notice duplication of: message with name "raw message"- fixing

USSAMOH2 (changes requested):

USSAMOH2:
You'll also need to add an endpoint to the REST API in the query router so that end-users can use this.

query-service/src/psql.rs
MrDarthShoe: I've discussed that with @esavier and I think we made a conclusion that inside Postgres there will be only tables which contain two strings and user should know that query will result in this kind of response

MrDarthShoe:
> You'll also need to add an endpoint to the REST API in the query router so that end-users can use this.

This will be included in another PR from my teammate after we get this QS part done

query-service/src/psql.rs
Kononnable: Ok, table structure contains only two strings, but in sql you can even `select 1, 1, 1` and such raw query will result in three columns. I think the reason to introduce raw queries in sql database in the first place is to let it do much more then just get single key and value(probably use some json functions to prepare data on the server).

USSAMOH2 (approved):

query-service/src/psql.rs
MrDarthShoe: This should be fixed now- I return Vec<u8> in GRPC.
It is not pretty though... Maybe you guys can propose a better way of getting values from rows in tokio_postgres

USSAMOH2:
@MrDarthShoe wouldn't you want to return `Vec<Vec<u8>>` a.k.a. `repeated bytes` for arbitrary query if someone returns multiple rows?

MrDarthShoe:
> @MrDarthShoe wouldn't you want to return `Vec<Vec<u8>>` a.k.a. `repeated bytes` for arbitrary query if someone returns multiple rows?

@USSAMOH2  currently I made a convention, that splits multiple columns just by inserting '\t', and rows by inserting '\n' between data-bytes.
Vec<Vec<u8>> is also an option, but what with multiple columns? In that case columns would be split by tab and rows by next Vec.
I think that Vec<Vec<Vec<u8>>> (every row has vector of columns and every field in column has vector of bytes) isn't pretty solution too.
If we want to merge these conventions into Vec<Vec<u8>>, I can do it but I need a green light for that

smores56 (approved):

luke-biel:
> > @MrDarthShoe wouldn't you want to return `Vec<Vec<u8>>` a.k.a. `repeated bytes` for arbitrary query if someone returns multiple rows?
> 
> @USSAMOH2 currently I made a convention, that splits multiple columns just by inserting '\t', and rows by inserting '\n' between data-bytes.
> Vec<Vec> is also an option, but what with multiple columns? In that case columns would be split by tab and rows by next Vec.
> I think that Vec<Vec<Vec>> (every row has vector of columns and every field in column has vector of bytes) isn't pretty solution too.
> If we want to merge these conventions into Vec<Vec>, I can do it but I need a green light for that

Vec<Vec<String>> OR Json containing array of arrays of strings are good. Vec<u8> is good as long as it's json too. Right now we require client of QS or QR to write custom parser for raw queries (also, tabs in column value will break it). I don't think that's the way.

MrDarthShoe:
> > > @MrDarthShoe wouldn't you want to return `Vec<Vec<u8>>` a.k.a. `repeated bytes` for arbitrary query if someone returns multiple rows?
> > 
> > 
> > @USSAMOH2 currently I made a convention, that splits multiple columns just by inserting '\t', and rows by inserting '\n' between data-bytes.
> > Vec is also an option, but what with multiple columns? In that case columns would be split by tab and rows by next Vec.
> > I think that Vec<Vec> (every row has vector of columns and every field in column has vector of bytes) isn't pretty solution too.
> > If we want to merge these conventions into Vec, I can do it but I need a green light for that
> 
> Vec<Vec> OR Json containing array of arrays of strings are good. Vec<u8> is good as long as it's json too. Right now we require client of QS or QR to write custom parser for raw queries (also, tabs in column value will break it). I don't think that's the way.

Now I understand: json should be coded inside raw bytes instead my own encoding convention. I've updated code to do so

luke-biel (approved):

esavier (approved):

PR #61, opened 2020-12-01
feat(data-router): Add details to publish error.

frondeus:


data-router/src/main.rs
```diff
@@ -164,12 +164,17 @@ async fn get_schema_topic(
...
     if delivery_status.is_err() {
         error!(
-            "Fatal error, delivery status for message not received. {:?}",
-            delivery_status
+            r#"Fatal error, delivery status for message not received. 
+            Topic: {},
+            Key: {},
+            Payload len: {},
+            {:?}"#,
+            topic_name, key, payload_len, delivery_status
```
luke-biel, line 177: Because you used r## it will keep newlines and spaces in output logs. I'm not sure you intended that. In will make logs look weird.

luke-biel (changes requested):

data-router/src/main.rs
frondeus: Ah damn, right!

luke-biel (approved):

PR #62, opened 2020-12-01
Release 0.0.3

frondeus:


luke-biel (dismissed):

.github/workflows/testing.yml
```diff
@@ -41,55 +41,3 @@ jobs:
         with:
           command: test
           args: --workspace
-
```
USSAMOH2, line 44: Why get rid of the integration testing?

examples/deploy/docker-compose.yml
```diff
@@ -6,109 +6,150 @@
...
-  schema_registry1:
-    image: schema_registry:latest
+  schema_registry:
+    image: cdl-schema-registry:latest
     build:
       context: ../..
       dockerfile: Dockerfile
       args:
         - BIN=schema-registry
+        - ENV=DEV
```
USSAMOH2, line 56: Why use debug builds?

query-service/src/psql.rs
```diff
@@ -74,8 +77,8 @@ impl PsqlQuery {
     fn collect_id_payload_rows(rows: Vec<Row>) -> HashMap<String, Vec<u8>> {
         rows.into_iter()
             .map(|row| {
-                let object_id = row.get::<usize, String>(0);
-                let value = row.get::<usize, String>(1).into_bytes();
+                let object_id = row.get::<usize, Uuid>(0).to_string();
```
USSAMOH2, line 80: This seems like it is just slower than simply retrieving as a string. We don't get more validation here, as non-UUID values cause a panic here.

query-service/src/psql.rs
```diff
@@ -74,8 +77,8 @@ impl PsqlQuery {
...
+                let value = row.get::<usize, Value>(1).to_string().into_bytes();
```
USSAMOH2, line 81: Instead of to_string and into_bytes, you should use serde_json::to_vec

USSAMOH2 (dismissed):

luke-biel (approved):

luke-biel:
Shall we remove building on develop pushes?
@frondeus

PR #65, opened 2020-12-02
feat(command-service): add PSQL schema as an optional argument

frondeus:


Kononnable:
Helm charts not updated.

frondeus:
> Helm charts not updated.

Cause by default you don't need to update helm charts, right?

Kononnable (dismissed):
nvm, missed `default_value = "public"` part

luke-biel (approved):

esavier (approved):

PR #67, opened 2020-12-02
fix(command-service): Use format with strong schema validation.

frondeus:
Unfortunately psql doesn't support `$1` parameter in schema name position because then it cannot validate table columns if these are correct.

### DIFF
--- a/command-service/src/output/psql/error.rs
+++ b/command-service/src/output/psql/error.rs
@@ -8,4 +8,8 @@ pub enum Error {
     RecvDropped,
     #[error("Failed to parse url into postgres connection string `{0}`")]
     FailedToParseUrl(bb8_postgres::tokio_postgres::Error),
+    #[error(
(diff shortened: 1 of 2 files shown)


smores56 (approved):

Kononnable (commented):
I would skip schema validation to postgres, there are some more rules, with some quirks - some schema db/table/schema names needs to be escaped to work properly in sql statement etc.

As for parameters - yep, it will fail if parameter is put in position which prevents generation of query plan(unknown db,schema,table, index name etc.).

frondeus:
I agree, but I'd say some validation is still better than no validation.

frondeus:
Maybe I should call it `sanitize` cause my first pririty was to ensure noone do some kind SQL injection here.
I know it's almost impossible because it would have to be a devops or a developer, not a end user but still.

Kononnable (approved):

PR #74, opened 2020-12-05
chore: Ci speedup

Kononnable:
- Add cache to our CI tests(~3x speedup).
- Check lock file compatibility on each build 

Currently cache usage default storage(cloud). We can probably get under 2 minutes if we configure storage locally: change caching method or point CI to local cache server: https://github.com/actions/toolkit/blob/7ad5004f874ebbe42d119b785dacb4360a0ce8f8/packages/cache/src/internal/cacheHttpClient.ts#L33

@luke-biel: I see `deny.toml` file, but don't see any CI action which uses that file. Is that intentional, or have we removed too much last time we changed github actions?


.github/workflows/testing.yml
```diff
@@ -2,11 +2,16 @@ name: Test
...
 on:
   push:
-    branches: [ develop ]
+    branches:
+      - "develop"
   pull_request:
 
 env:
   CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: 0
```
frondeus, line 11: Why disabling incremental compilation?

Kononnable: - first time(clean) compilation is slower with incremental builds(which is not really an issue when caching)
- incremental compilation gives us a lot of artifacts(size) - pushing incremental build artifacts to the cloud and fetching it again on next build in most cases takes more time then building without incremental. This might be different if files are cached locally.

luke-biel:
> @luke-biel: I see `deny.toml` file, but don't see any CI action which uses that file. Is that intentional, or have we removed too much last time we changed github actions?
It wasn't set up in the first place.

Kononnable:
So we ran it manually before? It looks like something which should be automatized.

luke-biel:
Problem is, there's a lot of issues that would need to be checked before so, otherwise we'd just turn our pipeline red.
I've added it as a way to check licenses easily back then, I think it's even mentioned somewhere. There was no time to do anything more.

esavier (changes requested):
please fix the merge's name

luke-biel (approved):

esavier (approved):

PR #75, opened 2020-12-07
test: use testcontainers for acceptance testing

luke-biel:


tests/README.md
```diff
@@ -24,28 +24,9 @@ Supported env variables are:
...
-| service | env | example |
-|---|---|---|
-| PostgreSQL user login | POSTGRES_USERNAME | postgres |
-| ... password | POSTGRES_PASSWORD | 1234 |
-| ... host | POSTGRES_HOST | localhost |
-| ... port | POSTGRES_PORT | 5432 |
-| CDL database name | POSTGRES_DBNAME | postgres |
-| Kafka broker | KAFKA_BROKERS | localhost:9092 |
+### Docker and docker-compose
+We test using testcontainers python library, which spins up new docker environment for every test.
```
frondeus, line 28:
```suggestion
We test using [testcontainers python library](https://pypi.org/project/testcontainers/), which spins up new docker environment for every test.
```

frondeus (dismissed):

tests/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: '3'
```
Kononnable, line 1: Should we remove zookeeper and kafka from 'old' docker-compose file(in examples)?

tests/common/postgres.py
```diff
@@ -12,15 +22,17 @@ def fetch_data_table(db):
 
 def clear_data_table(db):
```
Kononnable, line 23: This function is no longer used?

Kononnable (commented):
Tbh. I'm not a fan of the idea. If it spins up services for each test it would make tests make tests execution much longer.
I've been working on projects with similar approach and had been kicked by that in the ...

E.g. https://app.circleci.com/pipelines/github/typeorm/typeorm/1978/workflows/c90413bc-7a2a-4a25-8461-637dd6b5f3bf/jobs/6143 - it's just recreating db schema on each test, re spawning whole db engines would make it a few times longer.

Anyway can you show up numbers: how long it takes (before and after) and how many tests where executed? We've got few test now so it's not a problem, but in the future it may become one.

tests/common/__init__.py
```diff
@@ -19,3 +26,46 @@ def retry_retrieve(fetch, expected_rows, retries=10, delay=6):
...
+        raise Exception('Failed it setup kafka topic')
+
+
+def ensure_postgres_database_exists(config):
+    set_up = False
+    for i in range(1, 12):
+        try:
+            db = connect_to_postgres(config)
+            curr = db.cursor()
+            curr.execute("CREATE SCHEMA cdl")
```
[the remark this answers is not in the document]
luke-biel: I agree, we have this in 3 places right now. I want to address this soon, but need to consider how we want to do it.

Plan is:
- move helm to `deployment` folder
- put `sql` subfolder within
- link/refer to that folder in all 3 places

But if we want to move contents of that folder out of our repo into separate one, this isn't as easy

Regardless, I'd prefer not to think too much about SQL duplication. Yet

tests/docker-compose.yml
luke-biel: They can stay there no problem. I think we'll be able to remove examplary deployment once e2e tests are present - they should make sufficient documentation for new users.

tests/common/postgres.py
luke-biel: Hmm, I think so, will rid of it

luke-biel:
> Tbh. I'm not a fan of the idea. If it spins up services for each test it would make tests make tests execution much longer.
> I've been working on projects with similar approach and had been kicked by that in the ...
> 
> E.g. https://app.circleci.com/pipelines/github/typeorm/typeorm/1978/workflows/c90413bc-7a2a-4a25-8461-637dd6b5f3bf/jobs/6143 - it's just recreating db schema on each test, re spawning whole db engines would make it a few times longer.
> 
> Anyway can you show up numbers: how long it takes (before and after) and how many tests where executed? We've got few test now so it's not a problem, but in the future it may become one.

I partially agree with you. I don't want these tests to cover our applications whole. I thought it may be possible, but you have a point, it's useless, it will take too much time.

However, I still think there's case where we should use these tests. We can keep amount of test cases per each application limited, only necessary to ensure that integration with Kafka/PSQL/Rabbit/you name it works. Then we can have it fail fast, not start whole helm backed k8s cluster and pinpoint exact issue without having to look into helm logs for what have failed.

As for times, this takes 1:20min, my previous attempt was taking twice as much due to error, so we cannot really compare them.

tests/README.md
luke-biel: changed

tests/common/postgres.py
Kononnable: It's still here :)

tests/docker-compose.yml
Kononnable: I think using as few `docker-compose` files as possible is the goal - just to not forget keeping those files in (sort of) sync.
Anyway, it's not a blocker.

Kononnable (approved):
Apart from two comments on previous discussions - LGTM.

tests/common/postgres.py
luke-biel: Shit

tests/docker-compose.yml
luke-biel: I agree, I wanted to avoid symlinking nor referring to too many '..' dirs in this. I will address this, hopefully very soon.

Kononnable (approved):

PR #77, opened 2020-12-09
chore: helm charts for victoria metrics

Kononnable:
Helm chart for Victoria Metrics:
- infrastructure
- command service

plus some cleanup


esavier (dismissed):

frondeus (dismissed):

frondeus (approved):

benchmarking/Cargo.toml
```diff
@@ -8,11 +8,11 @@ license = "Apache-2.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
luke-biel, line 8: Can't we delete this whole folder for time being?

helm/helm-infrastructure/Chart.yaml
```diff
@@ -32,3 +32,6 @@ dependencies:
 - name: postgresql
   version: "9.4.0"
   repository: "https://charts.bitnami.com/bitnami"
+- name: victoria-metrics-single
```
luke-biel, line 35: Can you remind me why we keep .tar files instead of running `helm dependency update`?

benchmarking/Cargo.toml
Kononnable: Technically we can. But why?
You don't expect us to do any performance testing any time soon and maintaining those 'tests' are unnecessary cost or what is the reasoning behind that?

helm/helm-infrastructure/Chart.yaml
Kononnable: I guess I saw it somewhere online (in the docs or somewhere else) and just go went with that approach. As for `helm dependency update` it upgrades dependencies, you probably meant `helm dependency build`

I don't know which approach is more common(keeping those tar files or not). I guess we should use the approach most projects use for those files.

USSAMOH2 (approved):

benchmarking/Cargo.toml
luke-biel: I don't expect us to run these tests very soon and I'm 100% sure they shouldn't be in top dir.
I also believe we may rethink benchmarks at some point, with automation and generic messaging system in mind.

Maybe that's for another thread. I don't mind this PR being merged with these changes.

helm/helm-infrastructure/Chart.yaml
luke-biel: Agreed, but I can't find any info on that topic.

frondeus: I thought `helm dependency update` just downloads tar files.

PR #79, opened 2020-12-09
feat(schema-registry): Check if kafka topic exists

frondeus:


schema-registry/src/error.rs
```diff
@@ -50,6 +54,32 @@ impl From<jsonschema::CompilationError> for RegistryError {
     }
 }
 
+impl From<utils::messaging_system::CommunicationError> for RegistryError {
+    fn from(error: utils::messaging_system::CommunicationError) -> Self {
+        Self::MQError(error.to_string())
+    }
+}
+
+impl From<RegistryError> for Status {
```
Kononnable, line 63: Maybe it's the way to go, but it looks ugly :P

schema-registry/src/rpc.rs
```diff
@@ -90,6 +94,15 @@ impl SchemaRegistry for SchemaRegistryImpl {
             kafka_topic: request.topic_name,
         };
 
+        if !self
+            .mq_metadata
+            .topic_exists(&new_schema.kafka_topic)
```
USSAMOH2, line 99: This should also be done when we update the schema topic.

utils/src/messaging_system/metadata_fetcher.rs
```diff
@@ -0,0 +1,42 @@
+use std::time::Duration;
+
+use anyhow::Context;
+use rdkafka::{producer::BaseProducer, ClientConfig};
+
+use super::CommunicationResult;
+
+pub enum MetadataFetcher {
```
USSAMOH2, line 8: I'm not sure if this needs to be generic, but you don't need to remove it if you don't want to.

USSAMOH2 (dismissed):

utils/src/messaging_system/metadata_fetcher.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub async fn new_kafka(brokers: &str) -> CommunicationResult<Self> {
+        let producer = ClientConfig::new()
+            .set("bootstrap.servers", &brokers)
+            .create()
+            .context("Metadata fetcher creation failed")?;
+
+        Ok(Self::Kafka { producer })
+    }
+
+    pub async fn topic_exists(&self, topic: &str) -> CommunicationResult<bool> {
```
Kononnable, line 22: If this is abstraction over messaging protocol we probably should use name like 'destination channel/pipe' or something similar - topic is related only to kafka.

utils/src/messaging_system/mod.rs
```diff
@@ -4,31 +4,41 @@ pub mod consumer;
...
 
+pub mod metadata_fetcher;
+
 #[derive(Debug, Error)]
 pub enum CommunicationError {
     #[error("Error during communication via message queue \"{0}\"")]
     InternalError(String),
+
+    #[error("Error during joining blocking task \"{0}\"")]
+    TokioError(String),
```
Kononnable, line 15: I would still say it's internal error - not related to application business logic.

utils/src/messaging_system/metadata_fetcher.rs
frondeus: But then why we are keeping `topic` inside of vertex :)?

utils/src/messaging_system/mod.rs
frondeus: Then i'd need to make an error message more generic cause its no longer `communication via message queue`

luke-biel: Hmm, i think this means more to rename Internal somehow as opposed to merging these two

frondeus: Now I'm thinking it's no longer `CommunicationError`

frondeus: Changed it to... `Error` :)

Kononnable (approved):

PR #80, opened 2020-12-10
Chore:(druid): Add TODO comment for Druid

Tesel:
Druid was supposed to be moved at timeseries query service, so i made it

### DIFF
--- a/query-service-ts/proto/query.proto
+++ b/query-service-ts/proto/query.proto
@@ -6,6 +6,11 @@ service Query {
      rpc QueryByRange(Range) returns (TimeSeries);
 }
 
+service QueryDruid {
+    rpc QueryMultiple (ObjectIds) returns (ValueMap);
+    rpc QueryBySchema (SchemaId) returns (ValueMap);
+}
+
 message Range {
     string object_id = 1;
     string start = 2;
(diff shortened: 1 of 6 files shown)


frondeus (approved):

query-service-ts/proto/query.proto
```diff
@@ -6,6 +6,11 @@ service Query {
      rpc QueryByRange(Range) returns (TimeSeries);
 }
 
+service QueryDruid {
+    rpc QueryMultiple (ObjectIds) returns (ValueMap);
+    rpc QueryBySchema (SchemaId) returns (ValueMap);
+}
+
```
luke-biel, line 13: This shouldn't be added, we want to have unified interface

luke-biel (changes requested):
Druid ain't that simple.
It is not turned on (cannot be used), and when integrating we need to make sure it and vm use same gRPC protocol.

Definitely remove lines added to proto.
You can either put work in integrating druid inte qs or leave it as is but adding comment for future us that druid is WIP and needs to be readjusted. I'm okay with you leaving it to us since it's not related directly to VM. But if you have spare time, adjusting would be of great help.

query-service/src/druid.rs
```diff
@@ -1,3 +1,5 @@
+// This service is not supported atm
```
Kononnable, line 1: Who will see that comment? I think using `unimplemented` or todo in place something is lacking would be a better solution. 
Unless that's comment just for you to not forget about something - in that case please make PR as draft for the time being.

Tesel: Removed

luke-biel:
I meant that we should move code but not integrate it, not just add comment, but well, let's not waste more time on this. I'm closing it.

PR #82, opened 2020-12-11
feat(query-router): route requests based on schema type

damszew:
Query-router will now:
- using **schema_id** check in `schema-registry`, type of schema and route request either to `query-service` or to `query-service-ts`
- additional parameters (such as range and step needed to query timeseries data) is passed to `query-router` in requests http body

Issue: #15 
BREAKING CHANGE: endpoint `GET /single/<Uuid>` will be changed to HTTP POST and require request body

### DIFF
--- a/query-router/api.yml
+++ b/query-router/api.yml
@@ -4,7 +4,7 @@ info:
   version: 0.1.0
 paths:
   /single/{object_id}:
-    get:
+    post:
       summary: Retrieve single object
       operationId: getSingleObject
       parameters:
--- a/query-router/api.yml
+++ b/query-router/api.yml
@@ -22,6 +22,14 @@ paths:
           schema:
             type: string
             example: "15251181-f749-42e0-b4a4-e4b3d90e990d"
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              oneOf:
+                - $ref: '#/components/schemas/Empty'
+                - $ref: '#/components/schemas/Range'
       responses:
         '200':
           description: >-
@@ -85,3 +93,17 @@ paths:
             application/json:
               schema:
                 type: object
+
+components:
+  schemas:
+    Empty:
+      type: object
+    Range:
+      type: object
+      properties:
+        from:
+          type: string
+        to:
+          type: string
+        step:
+          type: string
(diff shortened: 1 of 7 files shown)


query-router/src/handler.rs
```diff
@@ -47,16 +84,30 @@ pub async fn query_multiple(
 
 pub async fn query_by_schema(
     schema_id: Uuid,
-    cache: Arc<AddressCache>,
+    address_cache: Arc<AddressCache>,
+    schema_type_cache: Arc<SchemaTypeCache>,
```
Kononnable, line 88: It looks to me like both caches could be merged - they are used together, fetching both values from schema registry can be done in single request(values are related to same schema_id)

damszew: Right, in most cases they will be used together

query-router/api.yml
```diff
@@ -22,6 +22,14 @@ paths:
           schema:
             type: string
             example: "15251181-f749-42e0-b4a4-e4b3d90e990d"
+      requestBody:
```
Kononnable, line 25: >Semantic error at paths./single/{object_id}.get.requestBody
GET operations cannot have a requestBody.
Jump to line 25

https://editor.swagger.io/

damszew: I know but it was decided on meeting with @luke-biel and @esavier that we want to pass additional parameters in request body.
Nevertheless "Request body" part is correctly shown in generated documentation

luke-biel: @Kononnable how do you see us sending complex requests? I understand it that we need to use GET (we are retrieving data) and queries are complex (now we have 2 params and don't know how many in future)

Kononnable: And here we go again, tabs or spaces? :smile: 

I know some software use get with request body(e.g. kibana), other uses post requests. If we want to be compatible with OpenAPI Specification we can't use request body in get query - even if RFC 7231 doesn't forbid it. For RFC request body has no known specification(undefined). 
As for using GET - I don't think it's the only option. GET responses can be cached and guarantees idempotency by definition(specification). POST doesn't do that but id doesn't say it's not idempotent either - according to specs post can change data but it's not required to. That means that POST can be used in such cases(however results won't be cached automatically).

TL;DR: I would use POST method here, it's ok with both specs and using get with request body will block us if we want to introduce any form of automatic checks on this api spec file.

damszew: All right, I've changed this endpoint to POST

Kononnable (approved):

PR #87, opened 2020-12-14
refactor: change internal msg format

luke-biel:


utils/src/message_types.rs
```diff
@@ -1,21 +1,41 @@
...
+use std::time::{SystemTime, UNIX_EPOCH};
 use uuid::Uuid;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct CommandServiceInsertMessage<'a> {
+#[serde(rename_all = "camelCase")]
+pub struct BorrowedInsertMessage<'a> {
     pub object_id: Uuid,
     pub schema_id: Uuid,
+    #[serde(default = "default_timestamp")]
```
Kononnable, line 12: Will lead to logic error which are hard to find. This should be set explicitly.

examples/deploy/docker-compose.yml
```diff
@@ -134,7 +135,7 @@ services:
       POSTGRES_HOST: postgres
       POSTGRES_PORT: 5432
       POSTGRES_DBNAME: postgres
-      POSTGRES_SCHEMA: public
+      POSTGRES_SCHEMA: cdl
```
Kononnable, line 138: Was this intended or is it some change which is applicable only to your local environment(I have no idea if schema would be created automatically)?

command-service/src/output/victoria_metrics/mod.rs
```diff
@@ -151,14 +153,13 @@ mod tests {
 
         #[test_case("{}"                => matches Err(Error::EmptyFields))]
         #[test_case("{ \"y01\": {} }"   => matches Err(Error::InvalidFieldType))]
-        #[test_case("y00=32q,y01=14.0f" => matches Err(Error::DataIsNotValidJson(_)))]
```
Kononnable, line 154: Removing test case in refactor PR isn't a good idea.

Kononnable (changes requested):
What is the reasoning behind those changes? It's not clear to me why we should have both cases.

command-service/src/output/victoria_metrics/mod.rs
luke-biel: It's not applicable anymore. RawValue cannot be an invalid json, str could.

utils/src/message_types.rs
luke-biel: Then DataRouter will require timestamps. This is now shared between DR and CS. I could make it optional, or '-1' by default. I don't know however if it's better. What do you think?

examples/deploy/docker-compose.yml
luke-biel: It's intended. Scripts provision this schema, example got out of sync.

luke-biel:
> What is the reasoning behind those changes? It's not clear to me why we should have both cases.

I don't understand which `both cases` are you referring to. If it's Owned and Borrowed message - it was already there. CS had Owned message which was called GenericMessage. I replaced it with common construct, and removed one unnecessary allocation in the process.

utils/src/message_types.rs
USSAMOH2: Between those two alternatives, I'd think that making it into a `pub timestamp: Option<i64>` makes it safer.

command-service/src/report/mod.rs
```diff
@@ -1,37 +1,38 @@
...
         Ok(())
     }
 }
 
 impl ReportSender {
-    pub fn with_message_body(&self, msg: &GenericMessage) -> Box<dyn Reporter> {
+    pub fn with_message_body(self, msg: &BorrowedInsertMessage) -> Box<dyn Reporter> {
         match self {
             ReportSender::Full(config) => Box::new(FullReportSender {
                 producer: config.producer.clone(),
```
USSAMOH2, line 32: If this consumes `self`, then we probably don't need to `clone` these.

USSAMOH2 (changes requested):

utils/src/message_types.rs
Kononnable: >Then DataRouter will require timestamps. 

I say this shows need for another structure. Data router input format(client data) can't have this field - it's internal to CDL(in theory we could just overwrite its value but its very unclean). If we allow users to use it we would lock ourselves to current format of data(timestamp) which is likely to change in the future. After some time we would get questions on why something doesn't work. We don't support inserting historical data at this point and people would try to use it as such and probably broke their CDL data in some future version.

command-service/src/output/victoria_metrics/mod.rs
Kononnable: It's a refactor so errors/tests don't get obsolete. The only possibility is if this error was redundant in the first place(code could never be executed because of logic conditions, technically value could be just unwrapped).

Kononnable: Or the other way around - in this PR we skip this error because of unwraping
`data: serde_json::from_str(&self.data.get()).expect("RawValue wasn't valid json Value"),`

luke-biel:
```        let event: BorrowedInsertMessage =
            serde_json::from_str(json).map_err(Error::PayloadDeserializationFailed)?;
``` is a place where this error will be caught now.
This refactor removed one unnecessary conversion. Earlier we checked if json was ok on deserializaton and then on serialization. Propagating type `InternalMessage` and RawValue means it is checked only on deserialization, and RawValue guarantees it holds a valid json. `build_line_protocol` had `payload: [u8]` argument, now accepts `payload: RawValue`. Slice could contain errors (in theory), RawValue can't. This function cannot, literally, fail here. Test is obsolete, because function changed.

File: utils/src/message_types.rs (hunk 1 above)
Comment (luke-biel) [reply utils/src/message_types.rs:12]: Ok, this was my misconception about cdl's timestamp. This needs separate type for DataRouter, since DR isn't supposed to receive timestamps.

File: data-router/src/main.rs [hunk 5]
```diff
@@ -38,6 +39,14 @@ struct Config {
     #[structopt(long, env)]
     pub cache_capacity: usize,
 }
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DataRouterInsertMessage<'a> {
```
Kononnable, line 44: I think this should go back to utils. Even if it's used only here it would be easier to find common message types when they're in one place.

Kononnable (commented):
I don't like how FullReportSender is implemented(or at least its name compared to what it does), not sure if `BorrowedInsertMessage` and `OwnedInsertMessage` could be handled better(seems like that but don't see a concrete solution yet).
Anyway those issues should not block that PR.

Only the one more comment about moving `DataRouterInsertMessage` should be addressed, otherwise LGTM.

data-router/src/main.rs
luke-biel: done

luke-biel:
> I don't like how FullReportSender is implemented(or at least its name compared to what it does), not sure if `BorrowedInsertMessage` and `OwnedInsertMessage` could be handled better(seems like that but don't see a concrete solution yet).
> Anyway those issues should not block that PR.
> 
> Only the one more comment about moving `DataRouterInsertMessage` should be addressed, otherwise LGTM.

I introduced NullObjectPattern because I could't come up with anything better that would not allocate memory when Reporting is disabled (coz payload has to be copied, for DE). Maybe we can hide that behind a feature? I'm just not sure whether that's the way.

Kononnable (approved):

Issue #16, opened 2020-11-17
Query Service - add "RAW" path


luke-biel:
Query Service should be able to send raw query to known database, even if common api does not support it.

This task has only timeseries query service in mention.

Acceptance criteria:
* [ ] `query-service-ts` has to be able to accept `Raw` grpc query, that would be passed directly to database, without translation process
* [ ] raw queries are integration tested

esavier:
not sure, but i would assume that there will be required some kind of field that is able to tell app that this query is raw or not, i.e. {rawtype:true} or something like that

PR #90, opened 2020-12-14
feat(query-service-ts): Added raw queries support for VictoriaMetrics

MrDarthShoe:
#16

### DIFF
--- a/query-service-ts/src/victoria.rs
+++ b/query-service-ts/src/victoria.rs
@@ -4,6 +4,7 @@ use reqwest::Client;
 use rpc::query_service_ts::{
     query_service_ts_server::QueryServiceTs, Range, RawStatement, SchemaId, TimeSeries, ValueBytes,
 };
+use serde::Deserialize;
 use structopt::StructOpt;
 use tonic::{Request, Response, Status};
 
@@ -38,6 +39,13 @@ pub struct VictoriaQuery {
     addr: String,
 }
 
+#[derive(Deserialize)]
+struct RawStatementParameters {
+    method: String,
+    endpoint: String,
+    queries: Vec<(String, String)>,
+}
+
 impl VictoriaQuery {
     pub async fn load(config: VictoriaConfig) -> anyhow::Result<Self> {
         let pool = Pool::builder()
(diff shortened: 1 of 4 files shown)


query-service-ts/src/victoria.rs
```diff
@@ -118,4 +119,24 @@ impl Query for VictoriaQuery {
             timeseries: response,
         }))
     }
+
+    async fn query_raw(
+        &self,
+        request: Request<RawStatement>,
+    ) -> Result<Response<ValueBytes>, Status> {
+        let url = format!("{}{}", self.addr, request.into_inner().raw_statement);
+        let body: serde_json::Value = reqwest::get(&url)
```
frondeus, line 128: What about post requests? Victoriametrics supports POST methods for example in `/query_range` because URL encoding has some limits. Also I think Graphite Tags API is using POST methods.

MrDarthShoe: Aren't POST's only for inserting data in PromQL? If so, adding metrics should be through command-service (or even directly to VictoriaMetrics, without CDL), not query-service-ts as I understand. In normal query-service we "allow" users to make INSERT's, just because we don't want to filter raw queries

frondeus: Not really. Victoriametrics provides API for Prometheus - https://github.com/VictoriaMetrics/VictoriaMetrics#prometheus-querying-api-usage.
Now If you follow link to Prometheus API - https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries
you can see both `GET` and `POST` methods are supported for **querying**, and not only for inserting.

> You can URL-encode these parameters directly in the request body by using the POST method and Content-Type: application/x-www-form-urlencoded header. This is useful when specifying a large query that may breach server-side URL character limits.

luke-biel (changes requested):
Some documentation on usage of this is a must. Even an example of json body that raw queries are supposed to send will be ok. You can place it in README of this crate.

Also:
> @MrDarthShoe MrDarthShoe deleted a comment from frondeus 21 hours ago
> @MrDarthShoe MrDarthShoe deleted a comment from frondeus 21 hours ago
Don't delete comments, even if they are useless/obsolete. We have no idea what was there now.

MrDarthShoe:
> Some documentation on usage of this is a must. Even an example of json body that raw queries are supposed to send will be ok. You can place it in README of this crate.
> 
> Also:
> 
> > @MrDarthShoe MrDarthShoe deleted a comment from frondeus 21 hours ago
> > @MrDarthShoe MrDarthShoe deleted a comment from frondeus 21 hours ago
> 
> Don't delete comments, even if they are useless/obsolete. We have no idea what was there now.

I planned adding example in query-router openapi after router part will be done, but surely I can add it in readme here too

luke-biel:
> > Some documentation on usage of this is a must. Even an example of json body that raw queries are supposed to send will be ok. You can place it in README of this crate.
> > Also:
> > > @MrDarthShoe MrDarthShoe deleted a comment from frondeus 21 hours ago
> > > @MrDarthShoe MrDarthShoe deleted a comment from frondeus 21 hours ago
> > 
> > 
> > Don't delete comments, even if they are useless/obsolete. We have no idea what was there now.
> 
> I planned adding example in query-router openapi after router part will be done, but surely I can add it in readme here too (which is better as user can talk directly to QS without QR)

Ok, then it's fine.

luke-biel (approved):

frondeus (approved):

PR #92, opened 2020-12-14
Use gRPC for communication between the Data Router and the Command Service

USSAMOH2:
For clean-up, this PR moved all gRPC definitions into a crate called `rpc`.


cdl-cli/src/actions/schema.rs
```diff
@@ -34,22 +31,22 @@ pub async fn get_schema(
 
 pub async fn add_schema(
     schema_name: String,
-    topic: String,
+    insert_address: String,
```
frondeus, line 34: I don't know if `insert_address` is right name. It can be either:
* Kafka topic
* GRPC address (URI)
* **OR** (technically but not realistically) - rabbitMQ exchange

command-service/src/input/config.rs
```diff
@@ -1,15 +1,9 @@
...
-    pub group_id: String,
-    #[structopt(long = "kafka-input-brokers", env = "KAFKA_INPUT_BROKERS")]
-    /// Comma separated list of brokers (eg. host1:9092,host2:9092)
-    pub brokers: String,
-    #[structopt(long = "kafka-input-topic", env = "KAFKA_INPUT_TOPIC")]
-    pub topic: String,
-
+pub struct InputConfig {
+    #[structopt(long = "grpc-input-port", env = "GRPC_INPUT_PORT")]
+    pub port: u16,
```
frondeus, line 6: I don't think we should replace it but make it optional, but my PR covers it.

command-service/src/input/service.rs
```diff
@@ -54,37 +38,39 @@ impl<P: OutputPlugin> KafkaInput<P> {
...
-            self.task_limiter
-                .run(async move || {
-                    if let Err(err) = Self::handle_message(router, message).await {
-                        error!("Failed to handle message: {}", err);
-                        process::abort();
-                    }
-                })
-                .await;
-        }
+        self.task_limiter
```
frondeus, line 65: You don't wanna spawn new task for GRPC handling. RPC needs to respond with an error if we fail to handle message.
Please note, if we simply respond with `Empty` and then crash, message will be lost. Fortunately, again my PR is going to fix it :)

schema-registry/src/error.rs
```diff
@@ -16,8 +16,6 @@ pub enum RegistryError {
     DbError(indradb::Error),
     #[error("Unable to connect to sled database: {0}")]
     ConnectionError(indradb::Error),
-    #[error("No topic found named \"{0}\"")]
-    NoTopic(String),
```
frondeus, line 20: Now this whole feature for checking kafka topic is useless.. And if we are going to fully switch to GRPC, shouldn't we ping `insert_address` to validate if it exists?

cdl-cli/src/actions/schema.rs
Kononnable: I would say `communication_channel` or something similar(generic) is more appropriate.
Same goes for rpc methods in schema registry.

cdl-cli/src/actions/schema.rs
```diff
@@ -76,7 +73,7 @@ pub async fn get_schema_versions(schema_id: Uuid, registry_addr: String) -> anyh
         .into_inner()
         .versions
         .into_iter()
-        .map(|v| Version::parse(&v))
+        .map(|v: String| Version::parse(&v))
```
Kononnable, line 76: Why explicit type?

cdl-cli/src/args.rs
```diff
@@ -50,8 +50,8 @@ pub enum SchemaAction {
         version: Option<VersionReq>,
     },
 
-    /// Get a schema's kafka topic from the registry.
-    GetTopic {
+    /// Get a schema's insert address from the registry.
+    GetInsertAddress {
```
Kononnable, line 54: As above - address doesn't cover all of the cases.

command-service/src/input/config.rs
Kononnable: It's better to keep refactor(extracting rpc crate) and feature changes in separate PRs so I would ask to remove all of feature changing changes from here, create another PR for that if necessary.

Kononnable (changes requested):

USSAMOH2:
Closed in favor of https://github.com/epiphany-platform/CommonDataLayer/pull/91.

PR #96, opened 2020-12-15
Move gRPC Definitions to `rpc` Crate

USSAMOH2:


frondeus (approved):
LGTM

frondeus (approved):
Still LGTM

schema-registry/src/lib.rs
```diff
@@ -1,10 +1,9 @@
 #![feature(drain_filter)]
 
-use crate::rpc::schema::schema_registry_client::SchemaRegistryClient;
-use crate::rpc::schema::{Empty, Id, ValueToValidate};
+use ::rpc::schema_registry::{Empty, Id, ValueToValidate};
```
luke-biel, line 3: Why need for `::` before rpc in this file?

frondeus: Because schema-registry has module called `rpc` so `::` means `crate, not module`

schema-registry/src/lib.rs
```diff
@@ -63,14 +54,14 @@ pub async fn get_view_of_data(
 }
 
 pub async fn promote_to_master(storage_addr: String) -> Result<String, RegistryClientError> {
-    let mut client = connect_to_registry(storage_addr).await?;
+    let mut client = ::rpc::schema_registry::connect(storage_addr).await?;
```
Kononnable, line 57: I guess we could just import `connect`, doesn't have to use full name then.

PR #103, opened 2020-12-16
chore(rpc): add license to crate 

luke-biel:


Kononnable (approved):
I think this is more like a `chore` not `fix` - doesn't touch functionality.

PR #109, opened 2020-12-18
feat(api): GraphQL server

frondeus:
[Rendered readme](https://github.com/epiphany-platform/CommonDataLayer/tree/graphql-api/api)


api/src/schema.rs
```diff
@@ -0,0 +1,124 @@
...
+    fn try_from(i: i32) -> Result<Self> {
+        match i {
+            0 => Ok(Self::DocumentStorage),
+            1 => Ok(Self::Timeseries),
+            i => Err(Error::InvalidSchemaType(i)),
+        }
+    }
+}
+
+impl Into<i32> for SchemaType {
```
luke-biel, line 37: From<SchemaType> for i32
will implement into and from at the same time

examples/deploy/docker-compose.yml
```diff
@@ -64,9 +73,13 @@ services:
       KAFKA_GROUP_ID: "schema_registry"
       KAFKA_TOPICS: "cdl.schema_registry.internal"
       INPUT_PORT: "50101"
+      IMPORT_FILE: "/var/data/schema.json"
+      #EXPORT_DIR: "/var/data/exported"
```
luke-biel, line 77: I don't see it necessary to include it within compose

api/src/schema.rs
```diff
@@ -0,0 +1,124 @@
...
+    pub query_address: String,
+    pub schema_type: SchemaType,
+}
+
+#[derive(Debug, juniper::GraphQLEnum, Clone, Copy)]
+/// Schema type, describes what kind of query service and command service is going to be used, as timeseries databases are quite different than others.
+pub enum SchemaType {
+    DocumentStorage,
+    Timeseries,
+}
```
luke-biel, line 24: Isn't there any way to use SchemaType from `rpc` crate? I don't like this duplication at all.

api/src/schema.rs
frondeus: Ah, this way around.... sure!

api/src/schema.rs
frondeus: It would be a violation of orphan rule, because I cannot implement graphql trait for `SchemaType` from `rpc`. I know it's not the best solution but hey, at least this enum has only 4 lines of code.

luke-biel: I was thinking maybe pack it in newtype, but dunno

api/src/queries.rs
```diff
@@ -0,0 +1,368 @@
...
+            let schema_type: i32 = schema_type.into();
+            conn.update_schema_type(rpc::schema_registry::SchemaTypeUpdate {
+                id: id.to_string(),
+                schema_type,
+            })
+            .await
+            .map_err(rpc::error::registry_error)?;
+        }
+
+        drop(conn);
```
Kononnable, line 187: Why separate drop?

api/src/queries.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        let versions = conn
+            .get_schema_versions(rpc_id.clone())
+            .await
+            .map_err(rpc::error::registry_error)?
+            .into_inner()
+            .versions;
+
+        let mut definitions = vec![];
+        for version in versions {
```
Kononnable, line 236: This probably should be executed in parallel. However it's probably not going to be a bottleneck soon.

api/src/queries.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        Ok(views)
+    }
+}
+
+pub struct Query;
+
+#[graphql_object(context = Context)]
+impl Query {
+    //TODO: Remove Option<> - return error on missing id.
```
Kononnable, line 286: Left it for future changes or forgot about that and was planned to be done before PR gets merged?

api/src/queries.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        Ok(View {
+            id,
+            name: view.name,
+            expression: view.jmespath,
+        })
+    }
+}
+
+// TODO: Add GRPC route to schema_registry which takes schema for uuid.
```
Kononnable, line 335: Same as above

api/src/schema.rs
Kononnable: Both enum and traits impl can be done from same place - those traits(unless I'm missing something) are general, not related directly to api crate business logic. It might be a good idea to put schema type into utils crate(in future we can thing of `common` or similar name).

api/src/schema.rs
```diff
@@ -0,0 +1,124 @@
...
+
+impl From<SchemaType> for i32 {
+    fn from(st: SchemaType) -> Self {
+        match st {
+            SchemaType::DocumentStorage => 0,
+            SchemaType::Timeseries => 1,
+        }
+    }
+}
```
Kononnable, line 44: Specify value with enum. Current form is error prone.

Something similar to https://stackoverflow.com/questions/31358826/how-do-i-convert-an-enum-reference-to-a-number

api/src/context.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::{config::Config, error::Error};
+use rpc::query_service::query_service_client::QueryServiceClient;
+use rpc::schema_registry::schema_registry_client::SchemaRegistryClient;
+use rpc::tonic::transport::Channel;
+use tokio::sync::{Mutex, MutexGuard};
+
+pub struct Context {
+    config: Arc<Config>,
+    registry_conn: Mutex<Option<SchemaRegistryClient<Channel>>>,
+    query_router_conn: Mutex<Option<QueryServiceClient<Channel>>>,
```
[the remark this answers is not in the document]
Kononnable: I think we both those fields are not necessary at this point. We're not reusing connection in other places.

We probably will need to do it at some point, but it won't be that simple to implement. Connection pooling implementation needs to allow service scaling(simple connection reuse without connection invalidation doesn't do that), probably some circuit breaker feature(depends on the way we go). As an additional disadvantage - using mutex this way limits ability to process multiple requests at the same time.

api/src/schema.rs
frondeus: But the point is, this enum needs to have implemented GraphQLEnum trait via `derive`. Do you wanna add graphql dependency to the common utils crate?

api/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+thiserror   = "1.0"
+env_logger  = "0.8"
+log         = "0.4"
+rpc         = { path = "../rpc" }
+tokio       = { version = "0.2", default-features = false, features = ["sync", "rt-threaded", "macros"] }
+uuid        = { version = "0.8", features = ["v4", "serde"] }
+juniper     = { version = "0.15", features = ["uuid"] }
+juniper_warp            = "0.6"
+#dataloader              = "0.13"
+#async-trait             = "0.1"
```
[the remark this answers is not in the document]
frondeus: One of the problems with GraphQL is N+1 problems. We are making N+1 requests to schema repository instead of 2.
Data loader might solve this issue. Please keep in mind that this PR is MVP and the work is still in progress. I commented this dependency for now because it's not used **yet** but I am going to use it in next PR.

api/README.md
```diff
@@ -0,0 +1,60 @@
+# GraphQL API Server - Overview
+
+server which provides `/graphql` and `/graphiql` routes for CDL management.
+It is self-describing, interactive and easy to use way to manage your instance.
+
+# Roadmap
+
+This crate is under heavy work-in-progress. It may change a lot with breaking changes. Use wisely.
```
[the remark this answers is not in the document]
frondeus: We can remove this roadmap as soon as I finish v0.2

api/README.md
```diff
@@ -0,0 +1,60 @@
...
+* [x] Documentation comments
+
+## v0.2
+* Insert object to CDL
+* Retrieve object from CDL
+* Use `dataloader` to solve N+1 problem (performance)
+
+# Getting started
+
+You can deploy web api on local computer via `docker-compose`
```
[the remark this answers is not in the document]
frondeus: Yes, for developer builds. That's why I wrote "on local computer". I wrote this README in case you guys want to play with this feature to decide if this is the correct way and if you like it.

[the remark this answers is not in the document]
frondeus: I'd love to do helm chart in the future, but right now I see more important things to do.

api/src/context.rs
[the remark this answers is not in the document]
frondeus: I guess @USSAMOH2  can answer these questions better than especially that he wrote it, but:
> We are not reusing connections in other places.
This is not an excuse for not introducing it here.

Anyway, I agree about mutex and probably we don't need this pooling right now, here.
We need to remember this tool is rather a for management than for daily-use. I do not expect 10k requests per second, not even 100rps. So we don't have to worry that much about performance. As soon as it's not sluggish-like-jira.

I'd be happy if we could do performance gains later and do incremental development here. It doesn't have to be perfect from the beginning.

api/src/queries.rs
frondeus: This is for future changes. In order to do it I have to implement new RPC route in schema registry.

api/src/queries.rs
frondeus: That's the change I was referring above ;-) We should ask about one schema instead of pooling all of them and filtering. It is on my TODO list, I'm doing it right now, but it's not blocking us, that's why I prepared PR.
The API is usable without this RPC call, just slower. I am going to fix it but meanwhile, you can get familiarized with graphql and the gains it brings.

api/src/schema.rs
frondeus: Sure I  can  use it... but how exactly it is error prone?

api/src/context.rs
```diff
@@ -0,0 +1,72 @@
...
+    type Target = C;
+
+    fn deref(&self) -> &Self::Target {
+        self.conn.as_ref().unwrap()
+    }
+}
+
+impl<'c, C> DerefMut for Conn<'c, C> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.conn.as_mut().unwrap()
```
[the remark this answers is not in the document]
frondeus: I guess the intention is here that `Option` is always `Some`.
In theory we could replace `Option` with `MaybeUninit`, because that's exactly what we are doing here.

api/src/schema.rs
Kononnable: Ok, didn't see that one - I guess if there is a comment after derive I just ignore that :smile: 

So the only options I see here is to leave it as is or create enum with single value(imported enum). Alias types won't work(can't derive on them).

api/src/schema.rs
Kononnable: Those codes are used twice - you can e.g. mistype value in one place and make `SchemaType::Timeseries => 0,` from `SchemaType::Timeseries => 1,`. It's easy to do especially when adding new value means copy pasting last line. Such bugs are hard to find(unless clippy suggest something to make code simpler).

frondeus: But your suggestion from link does not solve `TryFrom` case which means this "magic numbers" (0, 1) would be still in two places. Am I wrong?

api/src/queries.rs
frondeus: I agree that can be avoided. Right now there is a drop, because conn has a mutex lock, and in following line I'm retrieving new mutex lock causing a deadlock. I'm going to change the following line and then remove the drop by reusing conn.

api/src/schema.rs
frondeus: IMHO I should leave it. Enum with a single value wouldn't solve a thing cause when you derive, all fields/variants have to implement trait anyway.

Kononnable (dismissed):

api/src/schema.rs
luke-biel: In theory you can cast from `u8` to `enum` somehow I guess.

There's `num::Frem/ToPrimitive` which derives enum automatically. We can try that.

frondeus: Not bad especially that we are implicitly using `num-derive` in `prost`

Kononnable (approved):

Kononnable (dismissed):

deployment/compose/docker-compose.cdl.yml
```diff
@@ -14,6 +14,15 @@ services:
...
 
+  victoria_metrics:
+    image: victoriametrics/victoria-metrics:latest
+    ports:
+      - "50105:50105"
+      - "8428:8428"
+    command: -opentsdbHTTPListenAddr :50105 -retentionPeriod 0
+    volumes:
+      - ./docker-volume/victoria_metrics/db:/victoria-metrics-data
+
```
luke-biel, line 25: You already have VM entry in `docker-compose.yml`

luke-biel (changes requested):

api/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## v0.2
+* Insert object to CDL
+* Retrieve object from CDL
+* Use `dataloader` to solve N+1 problem (performance)
+
+# Getting started
+
+You can deploy web api on local computer via `docker-compose`.
+For more see [documentation](deployment/compose/README.md).
```
luke-biel, line 34: Path has changed

deployment/compose/initial-schema.json
```diff
@@ -0,0 +1,52 @@
+{
```
luke-biel, line 1: I now noticed that this file is in compose root. Can you move it to `docker-volume/schema-registry/init'?

api/src/main.rs
```diff
@@ -0,0 +1,43 @@
...
+                "<html><h1>juniper_warp</h1><div>visit <a href=\"/graphiql\">/graphiql</a></html>"
+                    .to_string(),
+            )
+    });
+
+    let state = warp::any().map({
+        let config = config.clone();
+        move || Context::new(config.clone())
+    });
+    let graphql_filter = juniper_warp::make_graphql_filter(crate::queries::schema(), state.boxed());
```
luke-biel, line 32: We will need cors to query this from some frontend. I guess '*' will suffice now.

```
let cors = warp::cors()
    .allow_any_origin();

let state = warp::any().map({
    let config = config.clone();
    move || Context::new(config.clone())
});
let graphql_filter = juniper_warp::make_graphql_filter(crate::queries::schema(), state.boxed()).with(cors);
```

luke-biel:
I've added 2 different schema definitions via
```
mutation AddSchemaDefMutation($id: Uuid!) {
  addSchemaDefinition(schemaId:$id, newVersion:{
    version: "1.1.0",
    definition: "{\"test\": false, \"values\": true}"
  }) {
    definition
  }
}
```
(definition for 1st was `{}`)
and then queried registry via
```
query SchemaDefinitionQuery($id: Uuid!) {
  schema(id:$id) {
    id,
    definitions {
      version,
      definition,
    }
  }
}
```

result
```
{
  "data": {
    "schema": {
      "id": "2cfad3c7-411a-11eb-8000-000000000000",
      "definitions": [
        {
          "version": "1.1.0",
          "definition": "{\"test\":false,\"values\":true}"
        },
        {
          "version": "1.1.0",
          "definition": "{\"test\":false,\"values\":true}"
        },
        {
          "version": "1.1.0",
          "definition": "{\"test\":false,\"values\":true}"
        }
      ]
    }
  }
}
```

I expected there to be versions 1.0.0 and 1.0.1. Something's not okay

deployment/compose/initial-schema.json
frondeus: The point is that we don't have to use docker-volume for schema, so why using `docker-volume/..` path?

frondeus:
It is okay.
You queried for `version` and schema registry uses semver. When retrieving data it returns you latest compatible definition for given version.
You can check it by adding:
```
query SchemaDefinitionQuery($id: Uuid!) {
  schema(id:$id) {
    id,
    definitions {
      version,
      definition,
      queriedVersion // <- this one will tell you `1.0.1` and `1.0.0`
    }
  }
}
```
If this isn't the desired behavior then we need to change the schema registry, but IMO this is job for different PR.

deployment/compose/initial-schema.json
luke-biel: This path can be named whatever, the point is to keep all `init` in one place, to not make mess in root of `compose/`. We can create new dir named `setup` or something.

luke-biel:
> It is okay.
> You queried for `version` and schema registry uses semver. When retrieving data it returns you latest compatible definition for given version.
> You can check it by adding:
> 
> ```
> query SchemaDefinitionQuery($id: Uuid!) {
>   schema(id:$id) {
>     id,
>     definitions {
>       version,
>       definition,
>       queriedVersion // <- this one will tell you `1.0.1` and `1.0.0`
>     }
>   }
> }
> ```
> 
> If this isn't the desired behavior then we need to change the schema registry, but IMO this is job for different PR.

I believe this is strange behaviour. I expect it to return all versions, not one version 3 times.

frondeus:
Ah good old `=` constraint can do miracle :)

api/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+warp        = "0.2"
+structopt   = "0.3"
+thiserror   = "1.0"
+env_logger  = "0.8"
+log         = "0.4"
+rpc         = { path = "../rpc" }
+tokio       = { version = "0.2", default-features = false, features = ["sync", "rt-threaded", "macros"] }
+uuid        = { version = "0.8", features = ["v4", "serde"] }
+juniper     = { version = "0.15", features = ["uuid"] }
+juniper_warp            = "0.6"
```
luke-biel, line 24: It feels wrong that less indented deps are behind this one

luke-biel (dismissed):

Kononnable (commented):
No helmfile changes, otherwise LGTM

api/Cargo.toml
```diff
@@ -12,15 +12,18 @@ path = "src/main.rs"
...
+structopt               = "0.3"
+thiserror               = "1.0"
+env_logger              = "0.8"
+log                     = "0.4"
+num-traits              = "0.2.14"
+num-derive              = "0.3.3"
+juniper_warp            = "0.6"
+juniper     = { version = "0.15", features = ["uuid"] }
 tokio       = { version = "0.2", default-features = false, features = ["sync", "rt-threaded", "macros"] }
 uuid        = { version = "0.8", features = ["v4", "serde"] }
```
luke-biel:
```suggestion
# Crates.io
env_logger  = "0.8"
juniper     = { version = "0.15", features = ["uuid"] }
num-derive  = "0.3.3"
num-traits  = "0.2.14"
log         = "0.4"
structopt   = "0.3"
thiserror   = "1.0"
tokio       = { version = "0.2", default-features = false, features = ["sync", "rt-threaded", "macros"] }
uuid        = { version = "0.8", features = ["v4", "serde"] }
warp        = "0.2"
juniper_warp            = "0.6"
```

luke-biel (approved):

USSAMOH2 (approved):

PR #114, opened 2020-12-22
feat(tests): Add component tests for query-router

damszew:
Issue #15


esavier (approved):

tests/common/config.py
```diff
@@ -23,12 +23,13 @@ def to_dict(self):
...
 class VictoriaMetricsConfig:
-    def __init__(self, database_url="http://localhost:12345"):
+    def __init__(self, database_url="http://localhost:8428"):
         self.database_url = database_url
 
     def to_dict(self):
         return {
-            "VICTORIA_METRICS_OUTPUT_URL": self.database_url
+            "VICTORIA_METRICS_OUTPUT_URL": self.database_url,
+            "VICTORIA_QUERY_URL": self.database_url + '/api/v1',
```
Tesel, line 32: I know it's tempting to concatenate strings, but it's better to use 
`from urllib.parse import urljoin`
...
`urljoin(self.database_url, '/api/v1')`

tests/common/config.py
```diff
@@ -23,12 +23,13 @@ def to_dict(self):
 
 
 class VictoriaMetricsConfig:
-    def __init__(self, database_url="http://localhost:12345"):
```
Tesel, line 26: I'd prefer to stay with port `12345`,  in order to not interfere with other deployments that run on `8428`, what other people has to say?

tests/common/query_router.py
```diff
@@ -0,0 +1,34 @@
...
+        env.update(SCHEMA_REGISTRY_ADDR=self.schema_registry_addr)
+
+        self.svc = subprocess.Popen([EXE], env=env)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.svc.kill()
+
+    def query_get_single(self, schema_id, object_id, body):
+        return requests.post(f"http://localhost:{self.input_port}/single/{object_id}", body, headers={'SCHEMA_ID': schema_id})
```
Tesel, line 28: urljoin could be used here aswell

tests/query_router/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    with cdl_env('.', postgres_config=PostgresConfig()) as env:
+        with QueryRouter('1024', '50103', 'http://localhost:50101') as qr:
+            with SchemaRegistry(str(tmp_path),
+                                "master",
+                                "localhost:9093",
+                                "schema_registry",
+                                "cdl.schema_registry.internal",
+                                "50101"
+                                ) as sr:
+                yield env, qr, sr
```
Tesel, line 27:
```suggestion
    with cdl_env('.', postgres_config=PostgresConfig()) as env,
           QueryRouter('1024', '50103', 'http://localhost:50101') as qr,
           SchemaRegistry(str(tmp_path),
                                    "master",
                                    "localhost:9093",
                                    "schema_registry",
                                    "cdl.schema_registry.internal",
                                    "50101"
                                     ) as sr:
                yield env, qr, sr
```

Tesel: GH screwed up formatting :/

tests/common/config.py
damszew: For test to not interfere with other deployments, all services (eg. kafka, postgress) also would need custom ports assigned

tests/query_router/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def assert_json(lhs, rhs):
+    assert json.dumps(lhs, sort_keys=True) == json.dumps(rhs, sort_keys=True)
+
+
+@pytest.fixture
+def prepare_env(tmp_path):
+    with cdl_env('.', postgres_config=PostgresConfig()) as env:
+        with QueryRouter('1024', '50103', 'http://localhost:50101') as qr:
```
Kononnable, line 19: Hardcoded values(port numbers etc.) should we allow some configuration? Like in `/tests/common/config.py`

tests/common/config.py
damszew: ok

tests/common/query_router.py
damszew: why use `urljoin` here? Imo formatting is cleaner here

tests/query_router/__init__.py
damszew: I'll improve it

Kononnable (approved):

tests/common/config.py
luke-biel: I'm changing tests soon to use compose from examples (having duplicate composes is bad idea), so this would need to be other way around. Tests use default, custom deployment uses, well, custom port.

luke-biel (approved):
LGTM

PR #126, opened 2020-12-28
Add simple batch decomposition in DataRouter

esavier:
I would like to send multiple cdl messages inside one MQ message,
Messages should look like this:
```
(pseudocode)
[ 
  {id:1,schema:1,data:{}}
  {id:2,schema:1,data:{}}
  {id:3,schema:1,data:{}}
  {id:4,schema:2,data:{}}
  {id:5,schema:3,data:{}}
]
```
decomposition should in this case look (at this point) as 5 separate messages containing one message each, without array

### DIFF
--- a/data-router/src/main.rs
+++ b/data-router/src/main.rs
@@ -101,7 +101,7 @@ async fn handle_message(
     
 
         // Check if something is an array
-        let value : Value = serde_json::from_str(message.payload()?).unwrap();
+        let value : Value = serde_json::from_str(message.payload()?).context("Payload detection failed, message is not in json format")?;
         if value.is_array() {
             let arr_messages: Vec<DataRouterInputData> = serde_json::from_str(message.payload()?)?;
             for event in arr_messages.iter(){
--- a/data-router/src/main.rs
+++ b/data-router/src/main.rs
@@ -94,43 +94,25 @@ async fn handle_message(
 ) {
     // Returns an empty result async for sending message to kafka after serialization
     let result: anyhow::Result<()> = async {
-        let since_the_epoch = SystemTime::now()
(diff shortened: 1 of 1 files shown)


data-router/src/main.rs
```diff
@@ -91,22 +92,46 @@ async fn handle_message(
...
+                let data = CommandServiceInsertMessage {
+                    object_id: event.object_id,
+                    payload: event.data,
+                    schema_id: event.schema_id,
+                    timestamp: since_the_epoch.as_millis() as i64, // TODO: Change types?
+                };
+
+                let payload = serde_json::to_vec(&data).unwrap_or_default();
+                let key = data.object_id.to_string();
+                send_message(producer.as_ref(), &topic_name, &key, payload).await;
```
Kononnable, line 119: Can we extract this logic to some function and use it on both sides of `if` statement?

data-router/src/main.rs
```diff
@@ -91,22 +92,46 @@ async fn handle_message(
...
+            }
+        } else {
+            let event  = serde_json::from_str(message.payload()?).context("Payload deserialization failed")?;
```
Kononnable, line 122: Already done in line 104

data-router/src/main.rs
esavier: yes, doing it now

data-router/src/main.rs
esavier: making it look "nice" is too much of a hassle, without much gain, also i personally relly dont like enum variants (and later enum recognition) which is one of the ways to do it.

esavier:
starting from scratch will be easier than merging at this point

PR #127, opened 2020-12-29
feat: use sqlx in place of bb8-postgres

luke-biel:
In regards to schema deletion, I still need to get confirmation. Then we can remove 'draft' flag.


luke-biel:
Needs more insight, we can close PR for time being

PR #129, opened 2020-12-30
chore(rust-toolchain): bump rust version to 30-12-2020

luke-biel:


schema-registry/src/types/storage/vertices.rs
```diff
@@ -7,7 +7,7 @@ use uuid::Uuid;
 use crate::types::{extract_vertex_property, SchemaType};
 
 pub trait Vertex: Sized {
-    fn to_properties<'a>(self) -> Vec<(&'a str, Value)>;
+    fn into_properties<'a>(self) -> Vec<(&'a str, Value)>;
```
Kononnable, line 10: Doesn't look even remotly related to rustc version bump. 
Is it a planned refactor, or just something which got here by mistake?

Kononnable (approved):

schema-registry/src/types/storage/vertices.rs
luke-biel: Clippy is bundled with rust version and it usually updates along.
An error started appearing after this bump, where it stated that `to_smth` should take `&self` by reference.

PR #131, opened 2020-12-30
ci: add markdown link check workflow

luke-biel:
I think we can find this handy

### DIFF
--- a/.github/workflows/check-docs.yml
+++ b/.github/workflows/check-docs.yml
@@ -3,12 +3,10 @@ name: Check Markdown links
 on:
   pull_request:
     paths:
-      - "**/*.md"
+      - "**.md"
   push:
     branches:
       - develop
-    paths:
-      - "**/*.md"
 
 jobs:
   markdown-link-check:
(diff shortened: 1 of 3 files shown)


.github/workflows/check-docs.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1
+        with:
+          use-quiet-mode: 'yes'
+          use-verbose-mode: 'yes'
```
Kononnable, line 19: Using quiet and verbose at the same time looks weird

Kononnable (approved):

.github/workflows/check-docs.yml
luke-biel: Quiet means we get only errors in stdout.
Verbose means reports (in our case errors) display http code.

I know it looks weird but it's action proposed by author. I can remove quiet if you insist.

PR #138, opened 2021-01-04
feat: add initial frontend version

luke-biel:
Things that will require a separate task:
* extend `mustard` with support of SVG buttons (I wan't CSS colored images)
* make graphql server address configurable from UI

I think this is valid initial iteration.
It may be good if we can use frondend for a while and gather feedback what can be improved.


USSAMOH2:
I am not a fan of the current deployment strategy. A standard frontend website is served out of a folder and requires no long-running process to provide frontend content unless you are using templates, and we aren't _actually_ using templates here. We should move this to deploy without needing the server, and only have one crate in this PR.

USSAMOH2:
When designing a website, it's important to have a consistent look: we need a CSS framework. I tend to prefer lighter-weight ones to make design decisions simpler and load times shorter. I like [Spectre](https://picturepan2.github.io/spectre/index.html) from this list of [CSS Frameworks](https://www.mockplus.com/blog/post/css-framework).

frondeus:
> When designing a website, it's important to have a consistent look: we need a CSS framework. I tend to prefer lighter-weight ones to make design decisions simpler and load times shorter. I like [Mustard](https://kylelogue.github.io/mustard-ui/index.html) from this list of [CSS Frameworks](https://www.mockplus.com/blog/post/css-framework).

While yes, it is nice to have consistent look - lets focus on UX and usability first. A shell but nice website is still just a shell.

luke-biel:
> I am not a fan of the current deployment strategy. A standard frontend website is served out of a folder and requires no long-running process to provide frontend content unless you are using templates, and we aren't _actually_ using templates here. We should move this to deploy without needing the server, and only have one crate in this PR.

I'm thinking about this. It's leftover from previous approach, where I had no graphql service. I'm gonna look into this closer to finishing this version. If after all is implemented I'm still not using server in any shape or for I'm gonna delete it.



> When designing a website, it's important to have a consistent look: we need a CSS framework. I tend to prefer lighter-weight ones to make design decisions simpler and load times shorter. I like [Mustard](https://kylelogue.github.io/mustard-ui/index.html) from this list of [CSS Frameworks](https://www.mockplus.com/blog/post/css-framework).

Right now we have CSS framework. It's LESS, and it's compiled in runtime (there's a script in layout.hbs for that). When I'm done with functionality, I will look into LESS compilation, or review other options.

frondeus:
> If after all is implemented I'm still not using server in any shape or for I'm gonna delete it.
I'd suggest having at least some kind of docker image with Nginx so it would be easily deployed on a local machine.

luke-biel:
@dependabot rebase

frontend/src/components/schema_registry/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+
+pub use schema_registry_add_definition::SchemaRegistryAddDefinition;
+pub use schema_registry_add_schema::SchemaRegistryAddSchema;
+pub use schema_registry_edit::SchemaRegistryEdit;
+pub use schema_registry_history::SchemaRegistryHistory;
+pub use schema_registry_list::SchemaRegistryList;
+pub use schema_registry_view::SchemaRegistryView;
+
+pub struct SchemaRegistry {
+    page: Page, // TODO: Make it a stack
```
frondeus, line 21: What about this TODO? Is this for the future or just missing?

frontend/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+
+mod cdl_objects;
+mod components;
+mod context_bus;
+
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+lazy_static! {
+    pub static ref GRAPHQL_URL: Url = Url::parse("http://localhost:50106/graphql").unwrap();
```
frondeus, line 23: Thats worrysome, it shouldn't be hardcoded

frondeus (changes requested):

frondeus:
While I appreciate your work, it's an amazing piece of tech I wonder if WASM/Yew is really that nice, while you get nice static typing, it's all just... so verbose and hard to look at.

Believe me, I'm not a fan of js, but `React` with `hooks` is surprisingly pleasant to type and extend.

Of course, I don't require a change from Yew to another framework, plus Sam is working on his implementation, these are just my thoughts on that matter.

frontend/src/lib.rs
luke-biel: Well, only way to make it configurable right now is to create a separate page, with config, that saves this to a cookie. It's a lot of work, I skipped it for now since it's not crucial for our (as in CDL team) experience with frontend right now. But yeah, we have to add it as the very next thing. I would just prefer to do it separately, so others may use and add features to frontend in parallel.

luke-biel:
> While I appreciate your work, it's an amazing piece of tech I wonder if WASM/Yew is really that nice, while you get nice static typing, it's all just... so verbose and hard to look at.
> 
> Believe me, I'm not a fan of js, but `React` with `hooks` is surprisingly pleasant to type and extend.
> 
> Of course, I don't require a change from Yew to another framework, plus Sam is working on his implementation, these are just my thoughts on that matter.

There's `yew_form` and `yew_router`, which remove much of this verbosity. Yew router requires an http_server that can handle SPA and I couldn't find any that would point `index/foo/bar/` back to `index` (i guess nginx can do this, but I can't use it locally) As for forms, i just didn't include it.

frontend/src/components/schema_registry/mod.rs
luke-biel: For the future. I may either add a Vec here or use yew_router

frontend/src/lib.rs
frondeus: What about using relative url like `/graphql` and working with reverse proxy like nginx?

Kononnable (changes requested):
Blocking till we decide which approach should we use(so it won't be merged by mistake, without proper code review).

For me it's very verbose and I don't like keeping html hidden inside other files just feels bad. Like keeping html content deep inside js files.

luke-biel:
Closing in favor of TS

PR #141, opened 2021-01-07
Move crates to separate folders

frondeus:


Cargo.toml
```diff
@@ -1,19 +1,6 @@
...
-    "data-router",
-    "document-storage",
-    "db-shrinker-postgres",
-    "leader-elector",
-    "query-router",
-    "query-service",
-    "query-service-ts",
-    "rpc",
-    "schema-registry",
+    "crates/*"
```
USSAMOH2, line 3:
```suggestion
    "crates/api",
    "crates/blob-store",
    "crates/benchmarking",
    "crates/cdl-cli",
    "crates/command-service",
    "crates/data-router",
    "crates/document-storage",
    "crates/db-shrinker-postgres",
    "crates/leader-elector",
    "crates/query-router",
    "crates/query-service",
    "crates/query-service-ts",
    "crates/rpc",
    "crates/schema-registry",
```

USSAMOH2 (changes requested):

Cargo.toml
frondeus: Okay, but why? Do we want to enable/disable crates by explicitly exclude them?

frondeus: Because if you want to easily check what crates are included in workspace I'd first go to `crates/` and do `ls` than check `Cargo.toml`

luke-biel: Shouldn't we just move all cargo related stuff to `crates/` folder?

frondeus: I think workspace Cargo.toml/lock should be in the root folder.
Otherwise you won't be able to do `cargo build` from `docs/` folder etc.

luke-biel: Should you be able? And there's always `--manifest-path`

Kononnable (approved):
Remember to change PR title/squash commit message on merge

USSAMOH2 (approved):

PR #152, opened 2021-01-11
feat: casuality defined by user

Kononnable:
- [x] Data model - add optional orderGroupId field
- [x] Data router, Command service - communication via rabbitmq
- [x] Data router - processing messages one by one(to keep ordering)
- [x] Rabbitmq - persistent messages
- [x] Command service - listening on multiple queues
- [x]  Command service - process messages in proper order(not all, only those defined by casuality)
- [x] Helm definitions changes to make code work, add bindings to rabbitmq 
- [x] Helm definitions - fix logging
- [x] Code cleanup
- [x] Unit testing of the new task queue
- [ ] Proper manual testing
- [x] Docker compose changes
- [x] Make e2e tests to pass
- [x] Docs


crates/command-service/src/input/service/mesage_queue.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    pub async fn listen(self) -> Result<(), Error> {
+        let mut streams = Vec::new();
+        for consumer in self.consumer {
+            let consumer = consumer.leak(); // Limits ability to change queues CS is listening on without restarting whole service
+            let stream = consumer.consume().await;
+            streams.push(stream);
+        }
+        let message_stream = select_all(streams.into_iter().map(Box::pin));
```
luke-biel, line 116:
```
message_streams = [`fut1`]

next() => async handle_message(fut1.msg1)
next() => async handle_message(fut1.msg2)
// random delay on tokio task with fut1.msg1
fut1.msg2.ack()
fut1.msg1.ack()
```
What keeps the order here? I see we read messages from partitions in order, but even when tokio runs on single thread, we have no guarantees when tasks will finish, and in which order they finish.
Unless, of course, same future cannot be read from before acking previous message.

crates/command-service/src/input/service/mesage_queue.rs
```diff
@@ -0,0 +1,138 @@
...
+    task_queue: Arc<ParallelTaskQueue>,
+}
+
+impl<P: OutputPlugin> MessageQueueInput<P> {
+    pub async fn new(
+        config: MessageQueueConfig,
+        message_router: MessageRouter<P>,
+    ) -> Result<Self, Error> {
+        let mut consumers = Vec::new();
+        let ordered_options = Some(BasicConsumeOptions {
```
luke-biel, line 34: According to yesterday, this has to be kafka

crates/command-service/src/input/service/mesage_queue.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+        let message_stream = select_all(streams.into_iter().map(Box::pin));
+
+        pin!(message_stream);
+
+        while let Some(message) = message_stream.next().await {
+            let router = self.message_router.clone();
+
+            let task_queue = self.task_queue.clone();
+            self.task_limiter
```
luke-biel, line 124: Task limiter should be part of task_queue functionality. Here we are limiting how many tasks can be queued, not processed.
We may consider changing this `select_all` to `for loop` with one tokio task per stream - then only we are sure task-limiter from one queue won't block other queues.

crates/command-service/src/input/service/mesage_queue.rs
```diff
@@ -0,0 +1,138 @@
...
+        counter!("cdl.command-service.input-request", 1);
+        let message = message.map_err(Error::FailedReadingMessage)?;
+
+        let generic_message = Self::build_message(message.as_ref())?;
+
+        trace!("Received message {:?}", generic_message);
+
+        let _guard = generic_message
+            .order_group_id
+            .map(async move |x| task_queue.run_task(x.to_string()).await);
```
luke-biel, line 87: `x` is ugly name

crates/command-service/src/args.rs
```diff
@@ -58,21 +59,45 @@ impl Args {
...
+                    .ok_or(MissingConfigError("Connection string"))?;
+                let ordered_queue_names: Vec<_> = input_args
+                    .ordered_queue_names
                     .clone()
-                    .ok_or(MissingConfigError("Topic"))?,
-                task_limit: input_args.task_limit,
-            }),
+                    .unwrap_or_default()
+                    .split(',')
+                    .filter(|x| !x.is_empty())
```
luke-biel, line 76:
```suggestion
                    .filter(|queue_name| !queue_name.is_empty())
```

crates/command-service/src/args.rs
```diff
@@ -58,21 +59,45 @@ impl Args {
...
+                    .map(String::from)
+                    .collect();
+                let unordered_queue_names: Vec<_> = input_args
+                    .unordered_queue_names
+                    .clone()
+                    .unwrap_or_default()
+                    .split(',')
+                    .filter(|x| !x.is_empty())
```
luke-biel, line 84:
```suggestion
                    .filter(|queue_name| !queue_name.is_empty())
```

crates/command-service/src/input/service/mesage_queue.rs
luke-biel: Also, this does'n run any task, this locks a mutex. `acquire_permit()`, `guard()`, `task_permit()`, `acquire_task_guard()`, `acquire_guard_for_key()`?

crates/data-router/src/main.rs
```diff
@@ -104,11 +114,16 @@ async fn handle_message(
...
             schema_id: insert_message.schema_id,
+            order_group_id: insert_message.order_group_id,
             timestamp: current_timestamp(),
             data: insert_message.data,
         };
 
-        let key = payload.object_id.to_string();
+        let key = payload
+            .order_group_id
+            .map(|x| x.to_string().replace("-", "."))
```
luke-biel, line 124:
```suggestion
            .map(|tag| tag.to_string().replace("-", "."))
```

luke-biel, line 124:
```suggestion
            .map(|tag| tag.to_string().replace("-", "."))
```

crates/data-router/src/main.rs
```diff
@@ -104,11 +114,16 @@ async fn handle_message(
...
+            .unwrap_or_else(|| "unordered".to_string());
```
luke-biel, line 125: In CS you are using this to limit parallel processing. Shouldn't this be left out empty for unordered messages?

crates/command-service/src/input/service/mesage_queue.rs
frondeus: This has to be dynamic, user should choose via ENV variable/cmd attribute

crates/command-service/src/args.rs
```diff
@@ -23,20 +23,21 @@ pub struct Args {
 arg_enum! {
     #[derive(Clone, Debug)]
     pub enum IngestionMethod {
-        Kafka,
+        MessageQueue,
```
frondeus, line 26: Instead of replacing Kafka, we need to add RabbitMQ as a another option

crates/data-router/src/main.rs
```diff
@@ -53,6 +55,11 @@ async fn main() -> anyhow::Result<()> {
     )
     .await?;
     let producer = Arc::new(
+        CommonPublisher::new_rabbit(&config.rabbit_connection_string)
```
frondeus, line 58: Again producer should be dynamic based on ENV variable

frondeus: It could be a nice base for `GRPC` support in `Data Router - Command Service` edge

frondeus (changes requested):

crates/command-service/src/input/service/mesage_queue.rs
Kononnable: Agree, but I think it's better to have those unrelated changes in separate PR.

crates/command-service/src/args.rs
Kononnable: Agree, but I think it's better to have those unrelated changes in separate PR.

crates/data-router/src/main.rs
Kononnable: Agree, but I think it's better to have those unrelated changes in separate PR.

crates/command-service/src/input/service/mesage_queue.rs
Kononnable: I don't think task limiter should be merged with task queue. They cover different functionalities and can by combined easily(reusability).

>Here we are limiting how many tasks can be queued, not processed.

Define being processed - as I remember task limiter was introduced to avoid CS fetching to much data from queue and consuming resources(also slowing down in the process). If we check if messages have same order group id before task limiter we would go back to those problems.

>We may consider changing this select_all to for loop with one tokio task per stream - then only we are sure task-limiter from one queue won't block other queues.

I'm not sure if that's really a problem - If someone sees such problems often they should really look into how they're splitting their workflow - define different exchange-queue binding on rabbitmq or change the way order group id is generated in their system. Task limiter will slow down other queues only if almost all of its tasks have single order group id, so increasing `task_limit` is also a good way to migrate that. Proposed changes would probably increase this too, don't see a point in defining task limit per queue if it's just to limit resource ussage.

crates/command-service/src/input/service/mesage_queue.rs
Kononnable: this took me a while to figure out. This is just about renaming task_queue method name, right?

crates/data-router/src/main.rs
Kononnable: I think empty key may be disallowed in some message queues(or tools used for them). 
Should be extracted to some const anyway.

frondeus (approved):

crates/command-service/src/input/service/mesage_queue.rs
luke-biel: Yeah, sorry if that was ambiguous

crates/command-service/src/input/service/mesage_queue.rs
luke-biel: Consumers have to be Kafka to function on DE environment. That's what Jakub said. Am I missing something?

Kononnable: Deployment for specific client isn't related to ordering. Ordering should work for both rabbitmq and kafka. Currently we're supporting only one message queue at the time(changing MQ requires code changes). What I meant is I would prefer to this being merged as is and create separate PR introducing ability to change MQ based on passed params.

We can go back to kafka in this PR but I don't see a point of it - it have to be changed by other PR either way to support multiple MQ.

luke-biel: We need to deploy ordering supporting Kafka by the end of the month. As I see this PR doesn't provide this feature. I'm okay in doing this in separate PR but we are running out of time.

luke-biel (approved):

luke-biel (dismissed):

Kononnable:
Temporary disabled component tests - this PR should be merged just before merging PR(WIP) with support for both message queues which will enable them again.

luke-biel (approved):

luke-biel (approved):

PR #165, opened 2021-01-14
chore: move benchmarking outside crates, remove storages

luke-biel:


luke-biel:
Traces of blob-store and sleigh are left on branch `archive/blob-store` so it can be easily reverted.

USSAMOH2 (approved):

esavier (approved):

PR #166, opened 2021-01-14
feat(api): GraphQL subscriptions

frondeus:


esavier (dismissed):

crates/api/src/queries.rs
```diff
@@ -4,15 +4,67 @@ use crate::{
...
+
+#[graphql_subscription(context = Context)]
+impl Subscription {
+    async fn reports(context: &Context) -> ReportStream {
+        let topic = &context.config().report_topic;
+        let stream = context
+            .subscribe_on_kafka_topic(topic)
+            .await?
+            .try_filter_map(|ev| async move { Ok(ev.payload) })
+            .map_err(anyhow::Error::from)
```
USSAMOH2, line 32: Is there a way to do this more concisely?

frondeus: I guess there is no point on having custom `Error` in Context if we are simply doing:
`anyhow::Error -> ContextError -> anyhow::Error`.
I could just return anyhow::Error in all context methods.

frondeus: Or remove `anyhow::Error` and do proper library-like error management.

crates/api/src/events.rs
```diff
@@ -0,0 +1,121 @@
...
+
+            tokio::pin!(stream);
+
+            while let Some(item) = stream.next().await {
+                if let Err(e) = sink.send(item) {
+                    log::error!("Couldn't send message to sink: {:?}", e);
+                    break;
+                }
+            }
+            log::warn!("Kafka stream has ended");
```
Kononnable, line 89: What does it mean in terms of business logic(when does it happen)? Or if it should not have happen maybe we should restart the service or do something else(e.g. to avoid leaking memory)?

crates/api/src/main.rs
```diff
@@ -32,20 +34,49 @@ async fn main() {
         })
         .with(cors.clone());
 
+    let context = Context::new(config.clone());
+
     let state = warp::any().map({
-        let config = config.clone();
-        move || Context::new(config.clone())
+        let context = context.clone();
+        move || context.clone()
```
Kononnable, line 41: Looks weird - two clones on the same value, can this be simplified in some way?

crates/api/src/main.rs
```diff
@@ -32,20 +34,49 @@ async fn main() {
...
+                    .map(|r| {
+                        if let Err(e) = r {
+                            log::error!("Websocket error: {}", e);
+                        }
+                    })
+                    .await
+                })
+            }
+        })
+        .map(|reply| warp::reply::with_header(reply, "Sec-WebSocket-Protocol", "graphql-ws"));
```
Kononnable, line 71: This header should probably be somewhere in consts(of a dependency).

crates/api/src/queries.rs
Kononnable: If this is going directly to user I think it's fine. If it's going to our reporting service(kafka topic at this point) it would have to be put in some structure anyway(same as errors in data router - we mentioned it on Thursday I think).

crates/api/src/events.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::pin::Pin;
+use thiserror::Error;
+use tokio::sync::broadcast;
+
+// TODO: Probably could be replaced by OwnedMessage from kafkard?
+/// Owned generic message received from kafka.
+#[derive(Clone, Debug)]
+pub struct KafkaEvent {
+    pub key: Option<String>,
+    pub payload: Option<String>,
```
Kononnable, line 24: Is it really optional?

Kononnable (changes requested):
I think using kafka directly is a blocker - what is the point of abstraction if we're not going to use it and won't be able to have same code for handling multiple message queues.

crates/api/src/context.rs
```diff
@@ -1,25 +1,59 @@
...
+use crate::{config::Config, error::Error, events::EventStream, events::EventSubscriber};
 use rpc::schema_registry::schema_registry_client::SchemaRegistryClient;
 use rpc::tonic::transport::Channel;
+use std::collections::HashMap;
+use tokio::sync::Mutex;
 
+#[derive(Clone)]
 pub struct Context {
     config: Arc<Config>,
+    //                               Topic, Event stream
```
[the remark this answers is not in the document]
frondeus: Fixed in metrics PR, can it wait?

crates/api/src/events.rs
frondeus: (After I fixed few things in metrics PR) - it means that for example connection to kafka broker has been closed.
I'm not sure what to do then to be honest. Restart could be nice or reconnection.
But this is in separated task/thread therefore probably we need some way to change `HashMap` from `Context`. It is possible, after all I used `Arc<Mutex>`, but it means `Subscriber` would be depending on `Context`.

crates/api/src/events.rs
frondeus: It is in kafka. You can send messages without payload.

crates/api/src/main.rs
frondeus: I tried, but unfortunately `||` is `FnMut` I think.

crates/api/src/events.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+impl EventSubscriber {
+    /// Connects to kafka and sends all messages to broadcast channel.
+    pub fn new(config: &KafkaConfig, topic: &str) -> Result<(Self, EventStream), anyhow::Error> {
+        let (tx, rx) = broadcast::channel(32);
+
+        log::debug!("Create new consumer for topic: {}", topic);
+
+        let consumer: StreamConsumer<DefaultConsumerContext> = ClientConfig::new()
```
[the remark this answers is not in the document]
frondeus: Do you mean from `utils` crate?

crates/api/Cargo.toml
```diff
@@ -23,7 +23,14 @@ num-traits  = "0.2.14"
...
 tokio       = { version = "0.2", default-features = false, features = ["sync", "rt-threaded", "macros"] }
 uuid        = { version = "0.8", features = ["v4", "serde"] }
 warp        = "0.2"
-juniper_warp            = "0.6"
+juniper_warp            = { version = "0.6", features = ["subscriptions"] }
+juniper_graphql_ws      = "0.2.1"
+rdkafka     = { version = "0.24", features = ["cmake-build"] }
+async-stream            = "0.3.0"
+serde       = { version = "1.0.119", features = ["derive"] }
+serde_json  = "1.0.61"
```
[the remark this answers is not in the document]
frondeus: I don't see alphabetic order in other crates, so why here? Unless it supposed to be alphabetic but no-one cares?

crates/api/src/events.rs
frondeus: Fixed with `FnOnce` callback called `on_close`

crates/api/src/events.rs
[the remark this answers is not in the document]
Kononnable: Yes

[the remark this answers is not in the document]
frondeus: Right now it's not possible. I will create an issue so we can touch this problem with RabbitMQ implementation.

crates/api/src/events.rs
```diff
@@ -0,0 +1,99 @@
...
+        F: FnOnce(String) -> Fut + Send + 'static,
+        Fut: Future<Output = ()>,
+    {
+        let (tx, rx) = broadcast::channel(32);
+
+        log::debug!("Create new consumer for topic: {}", topic);
+
+        let consumer =
+            CommonConsumer::new_kafka(&config.group_id, &config.brokers, &[topic]).await?;
+        let consumer = Box::leak(Box::new(consumer));
```
Kononnable, line 50:
```suggestion
        let consumer = consumer.leak();
```
Easier to trace(remove leaks once it's possible).

Kononnable (approved):
We probably should add an issue about leak which is actually dangerous here.
And another one about helm charts which is getting out of date(I think api isn't event there)

crates/api/src/events.rs
luke-biel: I say more, we can't leak here, every new subscription will allocate new Consumer that will live forever

crates/api/src/context.rs
```diff
@@ -1,25 +1,63 @@
...
 
-use crate::{config::Config, error::Error};
+use crate::{config::Config, events::EventStream, events::EventSubscriber};
+use anyhow::Result;
 use rpc::schema_registry::schema_registry_client::SchemaRegistryClient;
 use rpc::tonic::transport::Channel;
+use std::collections::HashMap;
+use tokio::sync::Mutex;
 
+type Topic = String;
```
luke-biel, line 10: This is misleading, either force use of newtype via tuple struct or use String everywhere.

luke-biel (changes requested):

crates/api/src/events.rs
frondeus: image
You dont have to worry about `leak` if there is no `leak` ;-)
`Subscriber` architecture uses `broadcast` which allowed me to remove the leak and have a normal local variable.

frondeus:
>I say more, we can't leak here, every new subscription will allocate new Consumer that will live forever

It was true only for two scenarios:
* You ask for new topic
* You lost connection and have to reconnect.
In our *notification* example it means there is usually only one leaked `Consumer` because we are **reusing** consumer between websocket clients.

And now this topic is obsolete - I just removed leak.

crates/api/src/context.rs
frondeus: I don't think `String` is an option. Then I need to tell via comment what `String` represents.
The whole
```
type Topic =  String;
```
acts as a documentation for whoever is reading code, what is the key of our hashmap.

frondeus: Especially that this type is not public and no method depends on it.

luke-biel: The issue here is that we don't name our other strings, nor document them. It's usually just parameter name, and if it's named well (eg. `topic`) it does same job, appears in same place, it enforced by compiler in same way.
We'd need to do same for other types for this to not be odd and out of place. But then I'd go with using newtype wherever possible.

luke-biel: In line 38 you have `topic: &str`.

frondeus: The fact that we aren't doing it for other strings is not a strong argument. We could improve our codebase one module at a time and it still is an improvement.

But whatever, I'm not going to dispute over silly `type Topic`, you win.

luke-biel (approved):

PR #171, opened 2021-01-15
WIP: feat(api): Simple metrics retrieval

frondeus:
Depends on #166


crates/api/src/kafka_events.rs
```diff
@@ -0,0 +1,80 @@
...
+                        .and_then(|s| std::str::from_utf8(s).ok())
+                        .map(|s| s.to_string());
+                    let payload = msg
+                        .payload()
+                        .and_then(|s| std::str::from_utf8(s).ok())
+                        .map(|s| s.to_string());
+                    KafkaEvent { key, payload }
+                });
+
+                tokio::pin!(stream);
```
frondeus, line 60: pin + while let could be in theory moved to common structure `Subscriber`

crates/api/src/config.rs
```diff
@@ -14,6 +14,18 @@ pub struct Config {
 
     #[structopt(long = "report-topic", env = "REPORT_TOPIC")]
     pub report_topic: String,
+
+    #[structopt(long = "metrics-interval-sec", env = "METRICS_INTERVAL_SEC")]
```
USSAMOH2, line 18: you don't have to specify the actual long/env names, you can just put `#[structopt(long, env)]` and it will use the appropriate casing.

crates/api/src/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+                    .then(|_| async { reqwest::get(&source).await })
+                    .and_then(|res| async move { res.text().await })
+                    .map_err(|e| e.to_string());
+
+                tokio::pin!(stream);
+                while let Some(item) = stream.next().await {
+                    // Error means there are no active receivers.
+                    // In that case we skip info and wait patiently until someone arrives
+                    // Therefore `let _`
+                    let _ = sink.send(item);
```
USSAMOH2, line 28: You can also do
```suggestion
                    sink.send(item).ok();
```

crates/api/src/context.rs
```diff
@@ -33,22 +48,58 @@ impl Context {
...
             .map_err(|e| Error::KafkaClientError(format!("{:?}", e)))
     }
+}
 
-    async fn consume_kafka_topic_inner(&self, topic: &str) -> Result<EventStream, anyhow::Error> {
-        let mut event_map = self.kafka_events.lock().await;
-        match event_map.get(topic) {
-            Some(subscriber) => {
-                let stream = subscriber.subscribe();
+impl Context {
```
USSAMOH2, line 69: Why have 2 impl blocks?

USSAMOH2 (changes requested):

crates/api/src/context.rs
frondeus: I often do 2 impls to visually separate `pub fn` from private `fn`

crates/api/src/context.rs
```diff
@@ -33,22 +48,58 @@ impl Context {
...
+        source: MetricsSource,
+    ) -> Result<MetricsStream, anyhow::Error> {
+        let mut event_map = self.metrics_events.lock().await;
+        match event_map.get(&source) {
+            Some(subscriber) => Ok(subscriber.subscribe()),
+            None => {
+                let source_addr = match source {
+                    MetricsSource::SchemaRegistry => &self.config.registry_metrics,
+                    MetricsSource::DataRouter => &self.config.data_router_metrics,
+                    MetricsSource::PostgresCommand => &self.config.postgres_command_metrics,
```
Kononnable, line 81: Repositories can be added add-hoc and even create by the user, so I'm not sure if we can use enum here.

crates/api/src/subscriber.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// We are using Box<dyn> approach (recommended) by Tokio maintainers,
+// as unfortunately `broadcast::Receiver` doesn't implement `Stream` trait,
+// and it is hard to achieve it without major refactor. Therefore we are using `async_stream` as a loophole.
+pub struct SubscriberStream<T, E>
+where
+    E: Debug + Display + Clone + Unpin + Send + Sync + 'static,
+    T: Send + Sync + Clone + Unpin + Send + Sync + 'static,
+{
```
Kononnable, line 37: Clean it up

crates/api/src/subscriber.rs
```diff
@@ -0,0 +1,107 @@
...
+    inner: Pin<Box<dyn Stream<Item = Result<T, SubscriberError<E>>> + Send + Sync>>,
+}
+
+impl<T, E> Subscriber<T, E>
+where
+    E: Debug + Display + Clone + Unpin + Send + Sync + 'static,
+    T: Send + Sync + Clone + Unpin + Send + Sync + 'static,
```
Kononnable, line 44: as above

crates/api/src/subscriber.rs
```diff
@@ -0,0 +1,107 @@
...
+{
+    pub fn new<F, S>(
+        name: &'static str,
+        mut consume: F,
+    ) -> Result<(Self, SubscriberStream<T, E>), anyhow::Error>
+    where
+        F: FnMut() -> Result<S, anyhow::Error>,
+        S: Stream<Item = Result<T, E>> + Send + 'static,
+    {
+        let (tx, rx) = broadcast::channel(32);
```
Kononnable, line 54: magic number

crates/api/src/metrics.rs
```diff
@@ -0,0 +1,35 @@
...
+
+pub struct MetricsSubscriber(Subscriber<Metrics, Error>);
+pub type MetricsStream = SubscriberStream<Metrics, Error>;
+
+impl MetricsSubscriber {
+    /// Connects to kafka and sends all messages to broadcast channel.
+    pub fn new(source: &str, interval: Duration) -> Result<(Self, MetricsStream), anyhow::Error> {
+        let (inner, stream) = Subscriber::new("metrics", move || {
+            let source = Arc::new(String::from(source));
+            let source = futures::stream::unfold(source, |s| async move { Some((s.clone(), s)) });
```
Kononnable, line 18: Definitely looks to complicated.

Kononnable (changes requested):
Please add a note in PR description if your PR is based on another PR - so it doesn't get merged to soon. And doi some cleanup before marking PR as ready to review...

crates/api/src/metrics.rs
frondeus: If you find anything less scary - be my guest, but unfortunately, that's the only way I know to pass owned string to stream.

frondeus:
> Please add a note in PR description if your PR is based on another PR - so it doesn't get merged to soon. And do some cleanup before marking PR as ready to review...

You imply that I didn't do a cleanup and that's not true.
I missed only the `Send + Sync` repetition.

crates/api/src/context.rs
frondeus: Okay, but then we need a clever way to pass metrics URL via env variable. I could parse and split by ",". The other issue is, right now we have separated dedicated query per each metrics source. We discussed before that probably graphQL API should parse metrics instead of passing raw string. And I agree but if we loose information from which component kind we are retrieving data, then how could we parse it and expect that `postgress-command` returns `countA` but `schema-registry` returns `countB` and `data-router` does not return counter at all?

PR #189, opened 2021-01-22
Dynamic queue choice based on ENV variables.

frondeus:
TODO:
* [x] HELM charts.
* [x] Refactor and cleanup


crates/schema-registry/src/main.rs
```diff
@@ -15,14 +17,46 @@ use std::path::PathBuf;
 use tonic::transport::Server;
 use utils::{metrics, status_endpoints};
 
+enum MessageQueueType {
+    Kafka,
+    Amqp,
+}
+
+impl<'de> Deserialize<'de> for MessageQueueType {
```
luke-biel, line 25: Hmm, can't you just derive Deserialize on the type? Serde supports renames

crates/api/src/events.rs
```diff
@@ -3,38 +3,37 @@
...
-use utils::messaging_system::consumer::CommonConsumer;
+use utils::messaging_system::consumer::{CommonConsumer, CommonConsumerConfig};
 use utils::messaging_system::Error;
 
-// TODO: Rename it to generic `Event` after adding support to RabbitMQ
-/// Owned generic message received from kafka.
+/// Owned generic message received from message queue.
 #[derive(Clone, Debug)]
-pub struct KafkaEvent {
+pub struct Event {
```
luke-biel, line 17: `Event` is a bit too generic for my taste.

crates/api/src/config.rs
```diff
@@ -1,33 +1,81 @@
...
+    #[structopt(long, env)]
+    kafka_group_id: Option<String>,
+    #[structopt(long, env)]
+    amqp_connection_string: Option<String>,
+    #[structopt(long, env)]
+    amqp_consumer_tag: Option<String>,
+}
+
+impl MessageQueueArgs {
+    pub fn config(&self) -> anyhow::Result<MessageQueueConfig> {
```
luke-biel, line 35: `into_config(self)` will skip you some clones

luke-biel (approved):

crates/schema-registry/src/main.rs
frondeus: It supports renames but here I deserialize with case-insensitive. I want to accept both `Kafka`, `kafka`, `KaFkA` etc. That's why I do `to_lowercase()`. And that's something serde doesn't support via an attribute.

luke-biel: Didn't notice `to_lowercase()`. You are correct.

crates/schema-registry/src/replication/mod.rs
```diff
@@ -121,18 +121,36 @@ impl ReplicationState {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct MessageQueueConfig {
+    pub queue: MessageQueue,
+    pub topic_or_queue: String,
+    pub topic_or_exchange: String,
```
Kononnable, line 128: Probably should change to something like 'output_topic_or_exchange' - we can distinct between those two fields only by rabbitmq distinction - name describes what it is technically not in context of business role.

Kononnable (commented):
LGTM(if helm files changes will be added)

luke-biel (approved):

deployment/helm/cdl/charts/druid-timeseries/templates/template.yaml
```diff
@@ -29,22 +29,27 @@ spec:
...
+        {{- else }}
+        - name: COMMUNICATION_METHOD
+          value: "amqp"
+        - name: AMQP_CONNECTION_STRING
+          value: {{ .Values.global.amqpConnectionString }}
+        - name: AMQP_CONSUMER_TAG
+          value: "cdl-druid-command-service"
+        {{- end }}
+        - name: ORDERED_TOPICS_OR_QUEUES
+          value: {{ .Values.inputTopicOrQueue }}
```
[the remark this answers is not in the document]
luke-biel: Druid uses `ts` field we supply in DR to "order" messages

PR #197, opened 2021-01-26
chore: move druid qs

### LINKED ISSUE #183: Refactor druid query service
Currently druid QS resides inside old QS crate, while it should be moved to new, QS-ts crate

luke-biel:
fixes #183


deployment/compose/README.md
```diff
@@ -87,3 +87,25 @@ eg.
...
+    coordinator \
+    broker \
+    historical \
+    router \
+    middlemanager \
+    schema_registry \
+    data_router \
+    druid_command \
+    druid_query \
+    query_router
```
frondeus, line 110: I understand why we pass so many services explicitly but it feels totally wrong and ridiculous. Cant we... I don't know, put it in a shell script or something?

deployment/compose/docker-compose.host-storage.yml
```diff
@@ -3,7 +3,7 @@
 # Please refer to README.md for instructions on how to run it.
 ###
 
-version: "3"
+version: "2.2"
```
frondeus, line 6: Why downgrade?

deployment/compose/README.md
luke-biel: This would be problematic, since we are to support 3 types of communication and N repos, we'd have to maintain N*3 scripts.
Using docker-compose alone you can setup all (for kafka in this case) when ommiting specyfying deps, or you can skip some when you explicitly tell what you want to set up.
I agree this is not the best case, but I don't think plain shell script will suffice. If I was to propose something, it would be helm wrapper, where you specify what repos to set up, what comms to use, and it would prepare k8s cluster for you.
But this is way out of scope of this task.

deployment/compose/docker-compose.host-storage.yml
luke-biel: Druid compose was using 2.2, we don't really use whatever 3 gives us, I wanted to keep consistency.

deployment/compose/README.md
frondeus: I agree it is not in the scope of this task.

frondeus: I will prepare a separate issue

frondeus (approved):

deployment/compose/docker-compose.host-storage.yml
Kononnable: version 2 is designed to work on single machines with docker
version 3 is designed to work with docker cloud
Both versions are 'alive', they're just meant to be used for different use cases(which is not so common knowledge unfortunately).

PR #200, opened 2021-01-27
test: adjust test environment variables

luke-biel:


tests/data/query_service_ts/schema/single.json
```diff
@@ -4,16 +4,16 @@
...
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=122 1608216912000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=123 1608216913000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=124 1608216914000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=125 1608216915000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=126 1608216916000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=127 1608216917000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=128 1608216918000000000",
-            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=129 1608216919000000000"
+            "a46a7fc0-0ca3-4852-affc-d589355524c5,object_id=6793227c-1b5a-413c-b310-1a86dc2d3c78 test=120 1611707410000000000",
```
luke-biel, line 7: This is a workaround, these tests need rewriting. Right now QueryService won't return data older than (i think) 30 days.

frondeus: Shouldn't we then just change query service? Or generate `single.json` instead of loading static file?

frondeus: Also, I think we can pass the workaround for now but we need a new issue which would describe this workaround and how to solve it properly

luke-biel: I intend to take care of it as soon as I'm done with druid, but yeah, gonna open an issue

frondeus (approved):

PR #201, opened 2021-01-27
Add Querying and Insertion to the GraphQL API

smores56:
Also, this fixes the build issue arising from a version mismatch of `k8s-openapi` between `kube` (indirect dependency) and `k8s-openapi` (direct dependency).


frondeus:
Can we split this PR into two - one with refactor of code structure (like moving to schema mod), etc, and another for actual feature changes? Then checking diff would be soo much easier

crates/api/src/schema/mutation.rs
```diff
@@ -148,4 +152,59 @@ impl Mutation {
...
+                &context.config().data_router_topic_or_queue,
+                &message.object_id.to_string(),
+                payload,
+            )
+            .await
+            .map_err(Error::PublisherError)?;
+        Ok(true)
+    }
+
+    async fn insert_transaction(
```
frondeus, line 182: is it really a transaction? What if 1st message passes but 2nd fails?

crates/api/src/schema/query.rs
```diff
@@ -157,4 +161,76 @@ impl Query {
         let mut conn = context.connect_to_registry().await?;
         get_view(&mut conn, id).await
     }
+
+    /// Return a single object from the query router
+    async fn object(object_id: Uuid, schema_id: Uuid, context: &Context) -> FieldResult<CdlObject> {
```
frondeus, line 166: Okay, thats nice and one way how to do it.
Another idea would be to add `object()` method to `Schema` query so you would do something like:
```
query {
    schema (id: "abcd") {
          object(id: "1234") {...}
    }
}
```

frondeus: Its not a requirement from me, just a tiny tip.

smores56: This would require loading the schema even when it's not needed, and I prefer the "separation of concerns" here of schemas from objects, so I'll leave it for now.

crates/api/src/schema/mutation.rs
smores56: changed to `insert_batch`

crates/api/src/schema/query.rs
frondeus: That's a good point.

frondeus (approved):

PR #210, opened 2021-01-29
feat(data-router): Message batching

frondeus:


frondeus:
Tested on docker-compose. Works with ordering.

crates/data-router/src/main.rs
```diff
@@ -212,6 +210,27 @@ async fn handle_message(
...
+        order_group_id: event.order_group_id,
+        schema_id: event.schema_id,
+        timestamp: current_timestamp(),
+        data: event.data,
+    };
+    let topic_name = get_schema_topic(&cache, payload.schema_id, &schema_registry_addr).await?;
+
+    let key = payload.object_id.to_string();
+    send_message(producer, &topic_name, &key, serde_json::to_vec(&payload)?).await;
+    counter!("cdl.data-router.input-single-msg", 1);
```
Kononnable, line 230: Probably duplicate with lines 161, 174, 184( last two of those lines lack of last dash).

crates/data-router/src/main.rs
```diff
@@ -251,8 +270,10 @@ async fn send_message(producer: &CommonPublisher, topic_name: &str, key: &str, p
             "Fatal error, delivery status for message not received.  Topic: `{}`, Key: `{}`, Payload len: `{}`, {:?}",
             topic_name, key, payload_len, delivery_status
         );
+        counter!("cdl.data-router.output-sendabort", 1);
```
Kononnable, line 273: this is useless:
1) service status should be fetched from k8s, not from service(which can misbehave)
2) this will never be sent to server - abort stops the process so no async operation will be done

crates/data-router/src/main.rs
```diff
@@ -251,8 +270,10 @@ async fn send_message(producer: &CommonPublisher, topic_name: &str, key: &str, p
...
         process::abort();
     };
+    counter!("cdl.data-router.output-singleok", 1);
```
Kononnable, line 276: not sure anyone will know what that metric does looking only on its name - maybe change it to something like schema registry requests(if we need it here, not on schema registry side)?

frondeus: But this counter has nothing to do with schema registry. Am I wrong?

Kononnable: I remember writing that regarding to counter increment made in `get_schema_topic`. Probably my bad - I guess there was some code rolled up by github and didn't check that's the start of another function.

Kononnable (approved):

PR #220, opened 2021-02-04
chore: Add extra tracing logs, delay exit

frondeus:


crates/command-service/src/input/service/message_queue.rs
```diff
@@ -88,6 +88,7 @@ impl<P: OutputPlugin> MessageQueueInput<P> {
 
     pub async fn listen(self) -> Result<(), Error> {
         let mut streams = Vec::new();
+        trace!("Consumers: {}", self.consumer.len());
```
smores56, line 91: "Number of consumers"?

frondeus:
```suggestion
        trace!("Number of consumers: {}", self.consumer.len());
```

smores56 (approved):

PR #222, opened 2021-02-04
fix(data-router): Make multitasking opt-out

frondeus:


luke-biel (approved):
Monotasking sounds weird. Maybe just `concurency`, `concurrent_processing`, `multitasking` and invert ifs?

smores56 (approved):

PR #228, opened 2021-02-05
feat(data-router): Add message version

frondeus:


luke-biel:
Let's have our API documented somewhere (a file with JSON-schema in the `docs` folder). Otherwise, this PR will be a silent change that will break their system once we bump the version - and it will break it without any notice, most likely.
Could you create a doc describing CDL API?

esavier (approved):

smores56 (approved):

luke-biel:
This is currently on hold, no point in keeping PR open, we can rebase when we decide to implement this again

PR #229, opened 2021-02-05
rfc: configuration service

luke-biel:
[rendered](https://github.com/epiphany-platform/CommonDataLayer/blob/rfc/configuration-service/docs/rfc/configuration_service.md)

Related thread: https://github.com/epiphany-platform/CommonDataLayer/discussions/230


docs/rfc/configuration_service.md
```diff
@@ -63,34 +63,59 @@ Schema-less environments (eg. no validator) still rely on schema-registry.
...
-
-All configuration in `configuration-service` should be backed by some database (we can consider sled here, however we should implemented 
-a wrapper over it, so features like replication are shared between `configuration-service` and `schema-registry`).
+On startup, all CDL services query `configuration service` for config using `CONFIGURATION_URL.` 
+Repository services would identify themselves using `repository_id.`
+`repository_id` is not unique and could be shared between multiple replicas of given service. 
+`command-service` and `query-service` belonging to one repository should share one `repository_id.`
+
+The final goal is to move most environment variables away from CDL services into a static,
+global database - `configuration service.`:
```
Kononnable, line 84: This final goal is outside of this rpc scope? As I recall original idea was just to move information about connecting to repository(where to pass data to CS, on which address QS is listening on). This scope however may change if we say so.

I'm not sure about some of those values. Some of them feels more like deployment config, not CDL related config - e.g. command service grpc port. It's not available to end users, in case of containerized environments it doesn't provide any value of its own(we don't have to worry about port collisions) - for such cases it could be a const as well.

esavier (commented):
I think its outdated ?

frondeus:
Decided to wait for resolution of #286

luke-biel:
Closing it right now. Branch will stay so we have a backup

PR #238, opened 2021-02-10
rfc: propose cdl input message versioning

luke-biel:
[rendered](https://github.com/epiphany-platform/CommonDataLayer/blob/rfc/cim-versioning/docs/rfc/protocol_version.md)

Related thread: https://github.com/epiphany-platform/CommonDataLayer/discussions/239

### DIFF
--- a/docs/rfc/CDLF-00010-00-rfc-01.md
+++ b/docs/rfc/CDLF-00010-00-rfc-01.md
@@ -96,6 +96,7 @@ in the second, only header with a too recent version.
 ## Decided solution
 
 We will follow with option 1, with an exception that `data-router` will check version of ingested message and fail if it's incompatible.
+Implementation will start when we will introduce major change to `CIM` format.
 
 # Further considerations
 ## Impact on other teams
--- a/docs/rfc/CDLF-00010-00-rfc-01.md
+++ b/docs/rfc/CDLF-00010-00-rfc-01.md
@@ -1,0 +1,110 @@
+# Front Matter
+
+```
+ Title: CDL Ingestion API versioning
+ Author: Łukasz Biel
+ Team: CDL
+ Reviewer: CDLTeam
+ Created on: 5/2/2021
+ Last updated: 10/2/2021
+ Tracking issue: https://github.com/epiphany-platform/CommonDataLayer/issues/225
+```
+
+# Introduction
+
+We need to introduce a way to version the `CIM` schema.
+
+## Summary
+
+We should provide JSON schema with `CIM` format.
+We need to return the most recent message version in case of deserialization failure in `DR`.
+
+## Glossary
+
+`CIM` - CDL Ingestion Message \
+`DR` - data router
+
+## Context
+
+Recent problems on the Digital Enterprise side have yielded it necessary
+to introduce `CIM` schema version tracking as a preliminary measure.
+
+## Goals and Requirements
+
+It should be transparent to CDL users which schema they have to use when inserting data into CDL.
+In case of version mismatch, `DR` must send a notification with information about the mishap, or,
+in the case of synchronous protocols, it must return a descriptive error.
+
+# Solutions
+
+## Existing solution
+
+`CIM` schema versioning does not exist.
+Users can find data format specifications in [data-router][data-router-readme].
+It contains only mandatory fields, skipping optional, e.g., `orderGroupId`.
+The most recent version of the data format is available when looking into the CDL code.
+
+## Proposed solution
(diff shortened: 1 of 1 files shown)


docs/rfc/protocol_version.md
```diff
@@ -0,0 +1,110 @@
...
+We need to return the most recent message version in case of deserialization failure in `DR`.
+
+## Glossary
+
+`CIM` - CDL Ingestion Message \
+`DR` - data router
+
+## Context
+
+Recent problems on the Digital Enterprise side have yielded it necessary
```
esavier, line 29: internal client, abstract it

docs/rfc/protocol_version.md
```diff
@@ -0,0 +1,110 @@
...
+* Let's assume we document a feature, e.g., new field `xyz`.
+  The user would send this field in payload expecting `feature` to work, but that depends on the `DR` he's running.
+  If the user is running an older version of `DR`, it would quietly discard the given field.
+  The whole premise of versioning is to ease debugging of errors encountered in CDL.
+  Using only `MAJOR` seems to go against it. We may need to add a `MINOR` to the mix, or report extra fields.
+
+### 2. Validate version, support multiple parsers
+
+We will publish specifications in the same way as mentioned in the prior solution.
+We will read the `version` field first (extracting it before parsing the whole JSON), and `DR` will choose the correct parser based on that version.
```
esavier, line 80: Does it mean that we will have to keep code matrix to handle each version separately?

docs/rfc/protocol_version.md
```diff
@@ -0,0 +1,110 @@
...
+Internally all messages would be parsed to one unified format.
+Such a feature will guarantee that the client can use new versions of CDL without upgrading his applications.
+In this case, we should assume that if a message has no `version` field, it uses **oldest**, probably `1.0` schema.
+
+#### Comments & Questions
+* How are we going to approach deprecations in this case?
+* Client still may send a too recent version of format to an old instance of CDL.
```
esavier, line 87: if we will use version field, it could be caught and warning/error can be issued, Versions have to be only forward compatible, and as stated before minor versions should only have optional fields added

docs/rfc/protocol_version.md
```diff
@@ -0,0 +1,110 @@
...
+
+#### Comments & Questions
+
+* Should we use JSON schema with MsgPack as well? In theory, we are only describing the format of `CIM`.
+  However, what if `MsgPack` requires more info that `JSONschema` would be able to convey?
+* Let's assume we document a feature, e.g., new field `xyz`.
+  The user would send this field in payload expecting `feature` to work, but that depends on the `DR` he's running.
+  If the user is running an older version of `DR`, it would quietly discard the given field.
+  The whole premise of versioning is to ease debugging of errors encountered in CDL.
+  Using only `MAJOR` seems to go against it. We may need to add a `MINOR` to the mix, or report extra fields.
```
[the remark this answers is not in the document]
frondeus: It is.

docs/rfc/protocol_version.md
luke-biel: Yeah, this is what 2nd proposal is about

luke-biel:
From discussion:

```

@Kononnable
Kononnable
on Feb 12
Maintainer

    Should we use JSON schema with MsgPack as well? In theory, we are only describing the format of CIM. However, what if MsgPack requires more info that JSONschema would be able to convey?

Same data which can be serialized in json is serializable in message pack(and other way around), so I don't think this is an issue.

    Let's assume we document a feature, e.g., new field xyz. The user would send this field in payload expecting feature to work, but that depends on the DR he's running. If the user is running an older version of DR, it would quietly discard the given field. The whole premise of versioning is to ease debugging of errors encountered in CDL. Using only MAJOR seems to go against it. We may need to add a MINOR to the mix, or report extra fields.

I'd prefer to throw an error instead of quietly discarding unknown fields. This way if old version of CDL is deployed, wchich is not really a CDL problem - someone who set up cluster did it wrong since required CDL version will be in user product requirements, request with 'new' field will result in an error.
```

Kononnable:
I think the important thing here is that right now we're not introducing any new field. We'll add version field when there is need for it(e.g. new required field). Otherwise LGTM.

luke-biel:
> I think the important thing here is that right now we're not introducing any new field. We'll add version field when there is need for it(e.g. new required field). Otherwise LGTM.

noted

Kononnable (approved):

PR #242, opened 2021-02-11
feat: custom kafka acknowledgments

### LINKED ISSUE #180: Custom kafka message acknowledgments
Consider implementing custom kafka message acknowledgment system.

Some of the messages send through CDL can be processed in parallel(messages with different order group id or with ordering feature disabled). If CDL services are communicating through kafka parallelization level is limited by number of kafka topic partitions. Sending ack on kafka message acknowledge all of the previous messages(changes partition offset) even if previous messages were not fully processed. To achieve durability in CDL(incoming message beeing processed at least once) we have to process messages one by one(within single partition) or use different acknowledgement strategy - acknowledge message only if all of the previous messages were processed successfully.

Note: List of messages processed, but not yet acknowledged should be capped. This will limit negative effects of the situation when message is lost and all of messages after it have to be processed twice(since none of them were acked).

Kononnable:
Closes #180 

Adds mechanism for custom kafka acknowledgments - it allows to process data in parallel with proper message acknowledging using  at-least-once guarantee. Kafka offset is committed to broker periodically.
Added task limiter to data router - same as in CS, it protects us from situation when we're processing to many messages at the same time(more time spend on context changes than on actual work).

Unfortunately due to kafka ability to dynamically change  partitions read by consumer(load balancing) we have to keep track of each message(otherwise it would be as simple as keeping track of three numbers). Alternative is to use separate stream per partition - it might be a bit faster solution(in terms of code execution), but will require much more code changes. This solution is worth considering if we encounter performance problems with custom acknowledgment system.

### DIFF
--- a/crates/data-router/src/main.rs
+++ b/crates/data-router/src/main.rs
@@ -96,7 +96,7 @@ async fn main() -> anyhow::Result<()> {
                 );
 
                 if !config.monotasking {
-                    task_limiter.run(move || async move { future.await}).await
+                    task_limiter.run(move || async move { future.await }).await
                 } else {
                     future.await;
                 }
--- a/crates/data-router/src/main.rs
+++ b/crates/data-router/src/main.rs
@@ -1,5 +1,3 @@
-#![feature(async_closure)]
-
 use anyhow::Context;
 use log::{debug, error, trace};
 use lru_cache::LruCache;
@@ -98,7 +96,7 @@ async fn main() -> anyhow::Result<()> {
                 );
 
                 if !config.monotasking {
-                    task_limiter.run(async move || future.await).await
+                    task_limiter.run(move || async move { future.await}).await
                 } else {
                     future.await;
                 }
--- a/crates/data-router/src/main.rs
+++ b/crates/data-router/src/main.rs
@@ -1,3 +1,5 @@
+#![feature(async_closure)]
+
 use anyhow::Context;
 use log::{debug, error, trace};
 use lru_cache::LruCache;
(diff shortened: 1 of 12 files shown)


crates/data-router/src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![feature(async_closure)]
```
frondeus, line 1: Why do you need this feature?

crates/data-router/src/main.rs
```diff
@@ -91,7 +98,7 @@ async fn main() -> anyhow::Result<()> {
                 );
 
                 if !config.monotasking {
-                    tokio::spawn(future);
+                    task_limiter.run(async move || future.await).await
```
frondeus, line 101:
```suggestion
                    task_limiter.run(move || async move { future.await }).await
```

crates/utils/src/messaging_system/kafka_ack_queue.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(Default)]
+pub struct KafkaAckQueue {
+    queue: Mutex<VecMap<KafkaPartitionAckQueue>>,
+}
+impl KafkaAckQueue {
+    pub fn add(&self, message: &BorrowedMessage) {
+        let partition = message.partition() as usize;
+        let mut queue = self.queue.lock().unwrap_or_else(abort_on_poison);
+        let partition_queue = match (*queue).get_mut(partition) {
```
frondeus, line 21: I think you could simplify this code by using `entry`, am I wrong?

frondeus (approved):

luke-biel (approved):

PR #248, opened 2021-02-12
feat: multithreaded data router with ordering enabled

Kononnable:
Enable multi threading even if message ordering is necessary.

Breaking change:
Requires message key to be same as order group id in case of ordered messages(or at least messages with same order group id should have the same key, doesn't have to be order group id value). For unordered messages it doesn't really matter - key can be empty or set to something concrete. It's important that key is not the same for every message(messages would be processed sequentially) which is the case for most workloads - otherwise kafka partitioning wouldn't work.

Let me know if you're ok with introducing those changes, I will make appropriate changes to documentation then.

EDIT:  Moved order_group_id from message payload to message key.


frondeus (approved):

luke-biel (approved):

frondeus (commented):
LGTM, resolve conflicts tho

Kononnable:
Added changes to docs, changed where order_group_id is stored(message key instead of message payload).

luke-biel (approved):

PR #250, opened 2021-02-12
fix(command-service): Properly build BorrowedInputMessage for GRPC

frondeus:


crates/command-service/src/input/error.rs
```diff
@@ -9,30 +8,30 @@ pub enum Error {
...
-    TimestampUnavailable,
     #[error("Failed to communicate with handler `{0}`")]
     CommunicationError(report::Error),
     #[error("Failed to initialize reporting module")]
     FailedToInitializeReporting(report::Error),
 }
+
+impl Error {
+    pub fn not_valid_group_id(error: uuid::Error) -> Self {
+        Self::NotValidUuid("Group ID", error)
```
Kononnable, line 29: I think using full name (order group id) would be more easier to read, we might introduce some other group field in the future.

Kononnable (approved):

Issue #76, opened 2020-12-08
Metrics port is not configurable


esavier:
port used in metrics is a const. set to 51805
i believe it should be taken from env and fallback to 51805 if not set otherwise

Kononnable:
What for? Metrics are internal to k8s(they don't go out of k8s), so no client is able to directly fetch them anyway.

Kononnable:
Resolved by #255

PR #255, opened 2021-02-15
test: organize black-box testing

luke-biel:
Tests depend on `docker-compose` changes that had to be included in this PR. Also, the metrics port had to be configurable.


crates/schema-registry/src/main.rs
```diff
@@ -91,7 +93,7 @@ pub async fn main() -> anyhow::Result<()> {
     };
 
     status_endpoints::serve();
-    metrics::serve();
+    metrics::serve(config.metrics_port.unwrap_or(58105));
```
Kononnable, line 96: It would be better if we could keep default values in one place(same place here and in other crates)

deployment/compose/README.md
```diff
@@ -30,82 +26,16 @@ Environment with infrastructure alone is started via:
...
-eg.
-
-* Adding new schema:
-> `cargo run -p cdl-cli -- --registry-addr "http://localhost:50101" schema add --name default-document`
+Adding new components happens through invoking `docker-compose` files containing parts of `CDL`.
+First, you must spin up `base` services, be it Kafka or RabbitMq, use:
 
-* Setting schema topic (in order for this schema to be routed to `command-service` topic must be `cdl.document.input`)
-> `cargo run -p cdl-cli -- --registry-addr "http://localhost:50101" schema set-topic --id 0a626bba-15ff-11eb-8004-000000000000 --topic "cdl.document.input"`
+`docker-compose -f docker-compose.cdl.(kafka|rabbitmq).base.yml up -d`
```
Kononnable, line 32: I would split it. I think code tags should contain only code which works out of the box/can be executed directly. Here, as I understand, reader needs to modify it first(remove kafka/rabbit with bracket). Same goes to entry below.

requirements.txt
```diff
@@ -1,8 +1,7 @@
...
 pytest~=6.1.2
-kafka-python~=2.0.2
-testcontainers~=3.1.0
-requests~=2.25.0
-grpcio~=1.33.2
-grpcio-tools~=1.33.2
-protobuf~=3.14.0
+kafka-python~=2.0.0
+grpcio~=1.35.0
+grpcio-tools~=1.35.0
```
Kononnable: This PR is supposed to remove testcontaiers, so I guess this file needs validation.

tests/common/query_router.py
```diff
@@ -16,9 +18,12 @@ def __enter__(self):
 
         env.update(CACHE_CAPACITY=self.cache_capacity)
         env.update(INPUT_PORT=self.input_port)
+        env.update(METRICS_PORT="59103")
```
Kononnable, line 21: Do we need to add another entry here on each new env variable?

tests/query_router/test_endpoint_raw.py
```diff
@@ -1,45 +1,99 @@
...
-    env, qr, sid = prepare_document_storage_env
+@pytest.fixture
+def prepare_postgres(tmp_path):
     data, expected = load_case('raw/query_ds', 'query_router')
 
-    db = connect_to_postgres(env.postgres_config)
-    insert_test_data(db, data['database_setup'])
-    db.close()
+    # declare environment
+    kafka_input_config = KafkaInputConfig("localhost:9092")
```
Kononnable, line 20: Why localhost? I mean, does it checks this value or is it just one that is not used but has to be specified.

Kononnable (commented):
I think this also closes #76 

One thing that is a hidden dependency of this change - docker-compose deployment is now a first-class citizen. It have to be up to date(or tests will fail). So far we're giving it as-is, without guarantees it will work without doing some manual work(because our officially supported environment is k8s). So we need to also test helm deployment files(so officially supported environment isn't tested less then unofficial one).

Second thing is that it's very verbose.

To sum up. I don't think those changes are bad. It's probably step in the right direction, but we're not on the end of the road yet.

And the most important part at the end - create PR descriptions when your PR covers multiple unrelated changes. Otherwise it's much harder for the reviewer to decide if some part of code  is necessary or is it just some bloat.

requirements.txt
luke-biel: I've noticed this oversight the moment you wrote a message on the channel. It's fixed

tests/query_router/test_endpoint_raw.py
luke-biel: It should be in defaults. That was by mistake, removed

tests/common/query_router.py
luke-biel: Leftovers from others, I simplified what I could here

deployment/compose/README.md
luke-biel: Good point

crates/schema-registry/src/main.rs
luke-biel: All I could do is create a const in utils. If we think about it - Option<u16> is not an option for a function argument, and utils shouldn't really provide "MetricsConfig" struct. So it's the only way I could think.

Kononnable (approved):

PR #260, opened 2021-02-17
rfc: Initial draft of the MessagePack RFC

### LINKED ISSUE #258: [RFC] CDLF-0000E-00 - Message Pack input
Research and provide the RFC with the findings on the possibility to use MessagePack as an input format for the CDL.

esavier:
* fixes #258


docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+
+##### Performance:
+* Comparing the performance to the other designs mentioned before, clearly shows that it is potentially more costly, ranging anywhere from `0` to `n*𝜅` where `n` is the number of formats CDL supports (currently this document describes the second format) assuming the worst scenario where `𝜅=m`
+
+##### Potential improvements:
+* assuming that one of the deserialization methods will fail, it is possible to parallelize those and to try to get at least one result out of several methods. This may improve performance to up to a single `𝜅`, while degrading memory usage (due to the possibility that deserialization methods will be destructive in the context of the used language) and requiring n threads to start working on the same message at the same time. This may prove helpful in cases when the user will want to use different formats and mix messages. Withal, it will be more troublesome to use for users that are using either one or the other.
```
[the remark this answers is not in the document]
Kononnable: I don't agree with `This may improve performance` - it may make processing single message faster, but will definitely not make system under load more performant, it will do the opposite(more work needs to be done per each message).

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+* depending on specification and partial loss of flexibility - DR will now employ another set of restrictions, however minor, to the message format to be able to correctly and concisely discern the incoming content format.
+
+
+## Further Considerations
+
+#### Impact on other teams
+Depending on the chosen solution, we may have to notify all users about the changes. It may also be necessary to introduce those as a "breaking change".
+
+#### Scalability
+Using Solution I may result in additional endpoints that have to be either provided or created. Apart from that, from initial research, there appear to be no impact on scalability whatsoever.
```
Kononnable, line 199: Don't think more endpoints is related to being able to scale in any way.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+* happy test 
+	proper behavior with proper values
+* format testing for each, readily available repository type
+	- document storage
+	- time series storage
+	- binary storage
+* performance testing with all the before mentioned cases
+
+#### Workload estimation
+Success rate should be acceptable, considering there is not a lot of solution-sepcific logic to introduce. Depending on which solution is chosen. It is initially roughly estimated to 20MD
```
Kononnable, line 219: Don't think we should estimate anything until specific solution is chosen. Estimates will vary between solutions.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+
+#### Out of scope
+Any other, not mentioned features, were not taken into consideration in this scope. Especially versioning (CDLF-00010-00)
+
+#### Open Questions:
+* Usage and behavior in Service mesh can not be checked at this point. This RFC should be revisited when the feature is complete and testing can be performed.
+
+### End Matter
+
+This is the version of the RFC that was deliberately cut down to investible size and format. The number of cases and solutions were condensed to those that were considered by the researcher the most valuable, leaving out minor variation that can be deduced from already provided solutions.
```
Kononnable, line 231:
```suggestion
```
Can we not introduce 'inner politics' to somewhat technical documents?

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+### End Matter
+
+This is the version of the RFC that was deliberately cut down to investible size and format. The number of cases and solutions were condensed to those that were considered by the researcher the most valuable, leaving out minor variation that can be deduced from already provided solutions. 
+
+#####  References:
+[MessagePack Specification](https://github.com/msgpack/msgpack/blob/3f0a9aae716596a86878c0d68dc0bd4256673202/spec.md)
+[CDL project](https://github.com/epiphany-platform/CommonDataLayer)
+[CDL - RFC discussions](https://github.com/epiphany-platform/CommonDataLayer/discussions/categories/rfc)
+[CDL - RFC candidates](https://github.com/epiphany-platform/CommonDataLayer/tree/develop/docs/rfc)
+[CDL - RFC releases](https://github.com/epiphany-platform/CommonDataLayer/tree/main/docs/rfc)
```
Kononnable, line 238:
```suggestion
```
It's common to all RFCs

Kononnable (commented):
Which solution is proposed? You've don some research, wrote it down, but I don't see anything indicating that you want to go one way or another.

Kononnable:
One thing that is lacking is how new format would be treated inside CDL - if it's only input and than inside CDL messages will be converted to json, or use both formats internally etc. Also what about output - do we want to change something there(binary data can still an issue there).

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+## Glossary
+
+##### Terminology
+* CDL - [Common Data Layer Project](https://github.com/epiphany-platform/CommonDataLayer)
+* DR - Data Router, an CDL component responsible for ingesting and routing initial messages.
+* SR - Schema Registry, an CDL component responsible for keeping information about the type of the object conveyed inside the message.
+* User - user of the CDL. In this case, the user knows how the CDL works, and is assumed to have access to the API and code unless stated otherwise.
+* Message - (abstract) message sent to CDL for processing
+* Breaking Change - change in behavior, or API of the CDL that may result with system breakage on release update.
+* MD - man-day - amount of work completed by one developer in one work day.
```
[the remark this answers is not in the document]
frondeus: I agree, it's too verbose for me.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+    Team            : CommonDataLayer
+    Reviewer        : CommonDataLayer
+    Created         : 2021-02-17
+    Last updated    : 2021-02-17
+    Version         : 1.0.12
+    CDL feature ID  : CDLF-0000E-00
+
+##### Abstract
+
+     The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
```
frondeus, line 14: How do we tackle the ambiguity of UPPERCASE vs lowercase? The RFC 2119 does not specify that.
* If we agree on UPPERCASE only then I see only 2 uses of these keywords.
* Otherwise, okay but we need to make it clear.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+* Breaking Change - change in behavior, or API of the CDL that may result with system breakage on release update.
+* MD - man-day - amount of work completed by one developer in one work day. 
+
+##### Features:
+[todo links]
+* CDLF-00004-00 - CDL Feature - Data Router introduction and message routing.
+* CDLF-00009-00 - CDL Feature - Message Batching - ability to digest array of proper CDL (v1 Input format) messages.
+* CDLF-0000A-00 - CDL Feature - Message Ordering - ability to guarantee linear message ingestion.
+* CDLF-0000C-00 - CDL Feature - GRPC as a main communication method, including DR input (direct ingestion)
+* CDLF-0000D-00 - CDL Feature - Introduction of Istio-based service mesh interoperability
```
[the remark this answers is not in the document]
frondeus: I don't understand why this section is necessary and what it brings to the table

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+* From initial evaluation it seems that this solution should not have any considerable performance drawbacks.
+* Parallel usage can result in performance degradation, however this is only a theoretical issue, and was not proven by testing yet due to pre-PoC state of the feature and no observed evidences of this behavior in other parts of  the CDL.
+
+##### Solution II - single endpoint with metadata carrier
+Both formats can use this same, abstract, transport endpoint. All the information required to control DR behavior and informing about the state of the message, will be provided in the transport's metadata. This means that we can retain the simplicity of having one, non-specific endpoint, where both message formats will be received.
+
+ After preliminary testing, it seems that all the protocols can pass metadata context alongside the payload.
+
+##### Notes:
+* For GRPC transport, information about the format would have to be contained in the header, or separate GRPC call (it is an open question, related to RFC on  CDLF-0000C-00 [todo link])
```
frondeus, line 133: Can we use reference style links for these `todo link`?

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+#### Background
+In present design, DR can only ingest JSON input format, which is both inefficient, and proves to be troublesome with some features. One example of which is being binary format ingestion. It was proposed to introduce MessagePack as an alternative. MessagePack is as flexible as JSON, meaning there is no need to recompile with each alteration in the protocol, and there is no need to utilise the message schema before initiating the communication. In addition, MessagePack allows the program to maintain binary data as is, without extra escaping or encoding. The Scope of this document include only information about the CDL input transport, which is, by extension, communication between user and DR.
+
+#### Assumptions
+We have to assume that some users may want to retain the ability to communicate over plain JSON, and/or choose a specific format for a specific job, meaning that two formats may have to be employed at all times.
+
+#### Preliminary testing
+
+##### Performance
+Simple preliminary testing using rust, C, and zig (while zig implementation being quite humble) displays some improvements to using message pack format over pure JSON string.
```
frondeus, line 70: While Zig performance is a nice addition, I would love to see the performance with more abstract and common languages like python and Java or JavaScript.

First of all Zig fills very similar niche to C and Rust, with zero-cost abstractions and being very close to bare metal. 

Second of all there is no runtime that might require a more sophisticated method to parse input stream (and to 
transform it into runtime object values).

Lastly - Zig is not a popular language and we should focus on the most common languages that might interop with CDL. Big business uses Java, C++, Python, and JS.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+Example serialization shows that the same message that was serialized in MessagePack was smaller than in JSON.
+| # | JSON    | message pack | difference | compression ratio
+|---|---------|--------------|------------|------------------|
+| # | 121 615 | 101 411      | 20 204      | 0.8338
+
+##### Limitations
+* integer values  are limited to 64 bytes 
+* maximum length of binary object is (2^32) -1 bytes (4294967295 bytes or around 3.9 GiB)
+* maximum length of string object is (2^32) -1
+* String may be malformed, or be a non-valid UTF-8 sequence
```
frondeus, line 92: Does it mean that the message pack does not protect consumers from a corrupted producer, right?

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+* maximum array length is (2^32) -1
+* maximum key-value map length is (2^32) -1 
+[Source: MessagePack Specification](https://github.com/msgpack/msgpack/blob/3f0a9aae716596a86878c0d68dc0bd4256673202/spec.md)
+ * During research, it was apparent that only a few rust libraries support full MessagePack features, for example extensions.
```
frondeus, line 96:
```suggestion
* During research, it was apparent that only a few rust libraries support full MessagePack features, for example extensions.
```

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+ * GRPC would have separate endpoint
+
+###### Notes:
+No testing nor PoC was created for this solution, please treat it as being theoretical.
+
+###### Major concerns:
+* This design will result in multiple configurations that have to be tested separately of each other and in tandem.  
+* Some amount of additional configuration have to be added.
+* Separate endpoints can be misused, resulting in a cascade of errors. Additional error handling has to be introduced to discern the type of issue, and whatever it was related to format errors or not.
+* Parallel usage of two or more endpoints may end up in thread starvation on constrained machines. In such example, by having one thread available for the workload, the focus would be shifted to one endpoint over another. This is an infrastructure related issue, but it has to be considered in this context.
```
frondeus, line 121: I would add that the fact about ordering guarantees. Because these messages would be on different topics/partitions/queues there is no guarantee (even inside one order_group_id) that messages will be processed in order.

luke-biel:
PR name should start with `rfc:`

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+
+##### Performance:
+* assuming the difference in metadata formats and our ability to both parse and serialize it, it can be safely assumed that performance will not suffer, however it depends on the way each library will get and use the metadata provided with the message. Comparing to the usual `O(m*m℘)` where `m` is the message size and `m℘` is the parsing cost, this solution will cost roughly `O((m*m℘) + (n*n℘))` where `n` and `n℘` is metadata length and its parsing cost respectively.
+
+##### Solution III - single endpoint with format recognition
+In this solution, both formats can be sent to one transport endpoint, similarly to the design proposed in Solution II, the message will not contain metadata in itself, but parsing will be performed for each available serialization method. In case all the methods fail, the message will be considered malformed and handled the usual way, which means reporting the error and continuing to work on the next queued message.
+
```
frondeus, line 147: More a comment than an issue: This is in parsing called "backtracking", you start parsing one variant, and if it fails you start parsing another.
It is considered rather bad practice.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+#### Scalability
+Using Solution I may result in additional endpoints that have to be either provided or created. Apart from that, from initial research, there appear to be no impact on scalability whatsoever.
+
+#### Availability problems
+Some of the solutions depends on the specific usage and/or implementation of the specific libraries and languge support. It is encouraged to put additional effort in weighting the specific solution against the other.
+
+#### Testing
+This feature MUST undergo thorough testing before being accepted. Test cases MUST include:
+*  failure scenarios:
```
frondeus, line 206:
```suggestion
* failure scenarios:
```

docs/rfc/CDLF-0000E-00-rfc-01.md
frondeus: IMHO estimations are connected to the implementation task, not to the RFC.

docs/rfc/CDLF-0000E-00-rfc-01.md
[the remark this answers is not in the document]
esavier: It is something that will be left in the repo for everyone to see. I am not commenting on your RFC's format, but i won't make mine something that i will be ashamed of later.

docs/rfc/CDLF-0000E-00-rfc-01.md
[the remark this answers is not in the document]
esavier: It links dependent rfcs and disclaims that those are related to this document at this stage.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+v1 Batch - it is a batch format introduced to alleviate some problems with transports. It conforms only to the v1 input format, and its treatment is described in RFC for CDLF-00009-00 [todo link]
+```
+[
+	{  v1 message }
+	{  v1 message }
+	{  v1 message }
+	...
+	=> n
+]
+ ```
```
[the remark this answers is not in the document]
esavier: batching as all of the CIM messages features, will have to be supported in messagepack, maybe i was unclear on that, i will add this to the specification

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+
+ After preliminary testing, it seems that all the protocols can pass metadata context alongside the payload.
+
+##### Notes:
+* For GRPC transport, information about the format would have to be contained in the header, or separate GRPC call (it is an open question, related to RFC on  CDLF-0000C-00 [todo link])
+* For Kafka and RabbitMQ, metadata can be stored in headers.
+* detached tests were executed, in which CDL context was not used, mostly focused on the availability and usability of the libraries, and it’s support. Additionally, PoC was not created for this solution, please treat it as being theoretical.
+
+##### Major concerns:
+* All the future transports will have to support sending some kind of metadata alongside the payload itself. This is quite possible but not guaranteed for all the users in all the use cases. In case new transport will be proposed, and it will not support it, it will have to be discarded or result in reopening this feature while choosing another solution.
```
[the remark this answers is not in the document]
esavier: You shouldn't start a complicated question with an "or" as it is confusing to read.
In case one transport will be able support metadata and the other are left with some kind of field, this way of approach will cause us to split message format to accommodate for this. An Example would be one transport using one set of fields and the other transport using a different set of fields.
This document was trimmed down from triple the size, I had to cut some material out. I assumed this is an obvious downside, but I can add this here if you wish.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+* Each transport have to get additional code, handling different ways of getting, and possibly parsing, the metadata provided with the message. 
+* It was also stated on the internal meetings that clients and/or libraries in different languages may have problems with this functionality, as it is not widely used. As per project directives, CDL have to be compatible with different languages, and this may be a potential issue.
```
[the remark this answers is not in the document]
esavier: Lapis had some issues in the rust side and I know that some C++ clients had problems with RabbitMQ, and GRPC as of November 2020, I unfortunately lacked the time to test everything and do research in the meanwhile.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -0,0 +1,240 @@
...
+
+##### Performance:
+* assuming the difference in metadata formats and our ability to both parse and serialize it, it can be safely assumed that performance will not suffer, however it depends on the way each library will get and use the metadata provided with the message. Comparing to the usual `O(m*m℘)` where `m` is the message size and `m℘` is the parsing cost, this solution will cost roughly `O((m*m℘) + (n*n℘))` where `n` and `n℘` is metadata length and its parsing cost respectively.
```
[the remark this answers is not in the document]
esavier: I think the explanation is clear, and is in the correct place. I will change it if you can point a valid and specific issue you have with it.

docs/rfc/CDLF-0000E-00-rfc-01.md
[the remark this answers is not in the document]
esavier: In this context, `this may` state that there is possibility for specific configurations to be faster if there are resources to spare. In normal situation, as stated afterwards, CPU and memory usage will degrade. I assumed that the RFC is quite clear on what it tries to convey when read carefully.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: This may be untrue to different transports that may be used in the future. This parts only states that we should be mindful about the necessity of having at least one endpoint per each supported format.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: That should be the case. Nevertheless, we did not have any rfc prior to planning. It is stated here that initially it was estimated to 20md, which by extension can change as a specific solution were to be chosen.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: I dont know what do you mean by `inner politics`. In case this fragment hurts your feelings in any way, this is, as you mentioned, technical specification and i will have to discard this suggestion, as it is not a part of the technical specification to keep someone happy.
This part says that the rfc was shortened, actually from 3times the size, for people to have better understanding without having to spend 2 hours reading it. It was requested on one of the meetings, which I complied with. One mistake i see right now is a spelling error which I will update.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: To be honest I do not know, I am open for suggestions.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: I do not know what the issue is?
Do you want to save space in the document? Is it some kind of compression method to yank out common parts of all of the RFC's?

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: I am not sure what do you mean.
Are you referring to `[text][reference]`? Yes, this was the initial idea. I wouldn't mind for each rfc to be named in some standard format as this one is. In that case, i wouldn't have to wait for rfc to be finished and could "assume" that when the rfc will be completed and in the proper branch, the link will work without adjustments.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: From my experience, C and C++ performance will be similar, especially that in case of optimizing for performance, C library can be used.
I tried languages that I know are fast to set up, and I am familiar with them enough to finish testing in around 2 hours per entry since time to submit the RFC was limited. If there is someone that would pick this issue up and provide test results for java or python I would include it in here.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: That is correct, at lest to my understanding of the specification. In our case. Technically, it may be a problem, since we may want to try and deserialize the message that is a valid message itself but will not be a valid UTF-8 string and thus potentially throw an error in rust.

esavier: I am unsure if this it is possible to force the library to deserialize string to vector, which would solve the issue. I will try to pick this up before another RFC review.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: Unfortunately, this is the part I cut out, but yes, this is how it will work. I will add a note about this issue before the next RFC review. Thank you.

docs/rfc/CDLF-0000E-00-rfc-01.md
esavier: Yes it is, i agree, but it is a major variation and I decided that it should have a place here. I tried to write it as objectively as possible without depicting anything as better or worse.

docs/rfc/CDLF-0000E-00-rfc-01.md
```diff
@@ -127,68 +128,77 @@ No testing nor PoC was created for this solution, please treat it as being theor
...
-* depending on specification and partial loss of flexibility - DR will now employ another set of restrictions, however minor, to the message format to be able to correctly and concisely discern the incoming content format.
+* Changes to existing v1 spec that are not clear or apparent. It means that v1 format will not change, but the way of delivery will. It is unclear right now how to announce those changes and if those should result in v2 format specification or not, also it is not known if the trimming should be used on the side of the receiver or the sender.
+* Depending on specification and partial loss of flexibility - DR will now employ another set of restrictions, however minor, to the message format to be able to correctly and concisely discern the incoming content format.
 
+##### Solution V - single endpoint with specific deserialization focus
```
Kononnable, line 193: How would it work with kafka? Only single consumer can read data from one partition.

smores56:
Presumably, this is out of scope for this issue, but I think gRPC is not the right format for intake into the data router. It provides a stable interface for inter-service communication and helps with message size, but something like HTTP works better if we are working with raw JSON and only one endpoint (both for client usablility: anyone can `curl` something, and for decoding: we can directly decode JSON instead of requiring the JSON be provided as a string and then decoded).

docs/rfc/CDLF-0000E-00-rfc-01.md
Kononnable: This could work when different types of data routers would use different consumer groups. However, there is still an issue about reporting malformed data. If message was not recognized by data router handling json messages it will though that it's message pack message. Same goes the other way. Problem rises when message is just corrupted - this situation should be reported.

luke-biel (approved):

Issue #186, opened 2021-01-20
Security vulnerability in dependencies: metrics-runtime->im->sized-chunks


MrDarthShoe:
**Describe the bug**
Cargo-deny advisories shouts that there is security vulnerability in dependencies:

```
error[A001]: Multiple soundness issues in Chunk and InlineArray
    ┌─ 
    │
326 │ sized-chunks 0.6.2 registry+https://github.com/rust-lang/crates.io-index
    │ ------------------------------------------------------------------------ security vulnerability detected
    │
    = ID: RUSTSEC-2020-0041
    = Advisory: https://rustsec.org/advisories/RUSTSEC-2020-0041
    = Chunk:
      
      * Array size is not checked when constructed with `unit()` and `pair()`.
      * Array size is not checked when constructed with `From<InlineArray<A, T>>`.
      * `Clone` and `insert_from` are not panic-safe; A panicking iterator causes memory safety issues with them.
      
      InlineArray:
      
      * Generates unaligned references for types with a large alignment requirement.
    = Announcement: https://github.com/bodil/sized-chunks/issues/11
    = Solution: No safe upgrade is available!
    = sized-chunks v0.6.2
      └── im v15.0.0
          └── metrics-runtime v0.13.1
              └── utils v0.1.0
                  ├── api v0.1.0
                  ├── command-service v0.1.0
                  ├── data-router v0.1.0
                  ├── query-router v0.1.0
                  ├── query-service v0.1.0
                  ├── query-service-ts v0.1.0
                  └── schema-registry v0.1.0
                      └── leader-elector v0.1.0

```

**Expected behavior**
No error

**To Reproduce**
```
cargo install cargo-deny
cargo deny check advisories
```
**Additional context**
We tried to use metrics-runtime v0.13.1 in project with cargo-deny in CI which was also failing.
This is root cause of problem: https://github.com/bodil/sized-chunks/issues/11

luke-biel:
I was aware of that, but there's not much we can do until `sized_chunks` update.
You may omit this dependency in `cargo deny` for time being

Kononnable:
Updating metrics will take care of the problem(dependency changes), but that's done in #137 - at least some part of upgrading metrics is related to tokio 1.0, not sure(didn't check it) if we can upgrade metrics to get rid of this problem without changing to tokio 1.0 first.

Kononnable:
Fixed long time ago.

PR #268, opened 2021-02-23
chore: testing version of rust-audit

esavier:


.github/workflows/rust-audit-check.yml
```diff
@@ -0,0 +1,14 @@
+name: Security audit
+on:
+  push:
+    paths: 
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
```
frondeus, line 6: I'd like to see this audit check also on pull request.

esavier: Added, ping, please resolve if you are satisfied :)

crates/data-router/Cargo.toml
```diff
@@ -5,8 +5,6 @@ version = "0.1.0"
 edition = "2018"
 license = "Apache-2.0"
 
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
smores56, line 8: If you're gonna remove this one, you should remove them from every crate. Currently, they are in every crate except for the `rpc` crate.

smores56 (changes requested):

crates/data-router/Cargo.toml
esavier: Technically, I removed it only to trigger the test, but sure, i don't mind

smores56: I'd say you should leave it (a.k.a. make another commit replacing this line), because it's useful to people reading `Cargo.toml` files unfamiliar with what is available.

.github/workflows/rust-audit-check.yml
```diff
@@ -0,0 +1,19 @@
+name: Security audit
+on:
```
Kononnable, line 2: It might be worth adding it on schedule for main branch/tags - so we know about security vulnerabilities in already released versions.

Kononnable (commented):
Currently it's returning four warnings, they're hardly readable though.
However according to #186 our dependency got a vulnerability. I think both cargo audit and cargo deny are using same RustSec advisory database.

crates/data-router/Cargo.toml
esavier: I already removed every occurrence though :(
I would argue if that's the case here to be honest. If someone is already familiar enough with Rust to tinker with CDL, they probably know where to find the documentation...?
Please state explicitly if I should revert the changes or leave it as is, I will not argue, and I don't really want to decide on this.

frondeus (dismissed):

.github/workflows/rust-audit-check.yml
esavier: I added periodic as discussed, it's a separate workflow which should trigger each day at 2:00 for both `main` and `develop`.
It should work, but there is no way of checking it out since scheduler is really wonky.

Please resolve this thread if you are satisfied with it.

frondeus (approved):

luke-biel (dismissed):
`cargo deny` can check rust security advisories, however, it checks only licenses
now. I would prefer altering `deny` action than adding a new tool.

smores56 (dismissed):

.github/workflows/rust-audit-check.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+    paths:
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
luke-biel, line 16: Why use `checkout@1` here when in other, and all our actions we use `@2`?

esavier: It's what was in the docs, but it shouldn't matter

esavier:
okay everything is fixed, please accept or lets drop this

esavier:
okay, requesting review again, lets roll with it

luke-biel (approved):

PR #275, opened 2021-02-25
chore: dockerfile simplification

Kononnable:


Dockerfile
```diff
@@ -11,53 +11,23 @@ RUN apt-get update && apt-get install -y cmake
...
-COPY crates/leader-elector/Cargo.toml crates/leader-elector/Cargo.toml
-COPY crates/query-router/Cargo.toml crates/query-router/Cargo.toml
-COPY crates/query-service/Cargo.toml crates/query-service/Cargo.toml
-COPY crates/query-service-ts/Cargo.toml crates/query-service-ts/Cargo.toml
-COPY crates/rpc/Cargo.toml crates/rpc/Cargo.toml
-COPY crates/schema-registry/Cargo.toml crates/schema-registry/Cargo.toml
-COPY crates/utils/Cargo.toml crates/utils/Cargo.toml
-COPY crates/api/Cargo.toml crates/api/Cargo.toml
-
-RUN cargo fetch
```
frondeus, line 31: By removing cargo fetch  you need to download all deps each time someone changes anything in the project, not only Cargo.toml/lock. Is that your intention?

Kononnable: Not in this case - notice lines 22 and 23. Cargo will fetch dependencies if needed, but if they're already downloaded it will reuse them.

Previous tries on this approach(like 3-6 months ago) ended badly because ... I was trying to cache the wrong path and didn't realize that.

frondeus (approved):

PR #276, opened 2021-02-25
test: helm deployment automated testing

Kononnable:
Automated tests which checks if our deployment files are up to date(for both kafka and rabbitmq)
Requires both #274 #275  to work

Results of previous builds:
https://github.com/epiphany-platform/CommonDataLayer/actions?query=branch%3Aci-k3s

Test action takes about 9-11 minutes when there is a cache hit, ~15 when there is none.


Cargo.ci.toml
```diff
@@ -0,0 +1,13 @@
+[workspace]
+members = [
+  "benchmarking",
+  "crates/*"
+]
+
+[profile.dev]
+debug = false
+incremental = false
```
Kononnable, line 9: I was unsure how to pass those values to CI based docker build. Other option would be add them to original file in CI step.

luke-biel:
```
cargo-features = ["named-profiles"]


[workspace]
members = [
  "benchmarking",
  "crates/*"
]

[profile.xyz]
inherits = "dev"
incremental = false

[profile.release]
lto = true
codegen-units = 1
```
We could create `ci` profile this way in normal `Cargo.toml`

> Note - this requires `-Z unstable-options` flag. But we can add it to `.cargo/Config.toml`. https://doc.rust-lang.org/cargo/reference/unstable.html

.github/workflows/testing-helm.yml
```diff
@@ -0,0 +1,86 @@
...
+jobs:
+  component-testing:
+    name: Test modules
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: install k8s
```
luke-biel, line 19: As I sidenote - thought about writing custom action for this?

Kononnable: Yeah, that idea came to my mind. I think the moment we need to use k8s deployment in our CI twice this is generally a thing to do. However, until them I'm fine with keeping it with other steps - it's easier to edit single file and don't have to learn specifics on github actions.

Cargo.ci.toml
Kononnable: Changed to named profiles. Tbh. I'm not a big fan of this feature(or rather its state) because custom profiles are only available to some commands, not to all of them, but in this case it suits rather perfectly.

luke-biel (approved):

PR #277, opened 2021-02-26
GRPC implementation

frondeus:
* [x] Command Service
* [x] API
* [x] Data Router
* [x] Schema Registry
* [x] Testing
* [x] Deployment files

I believe the core part of the implementation is done and won't change drastically therefore I decided to share this PR right now.


crates/api/src/events.rs
```diff
@@ -19,8 +22,10 @@ pub struct Event {
     pub payload: Option<String>,
 }
 
+type Sender<T = Event> = broadcast::Sender<T>;
```
Kononnable, line 25: You expect this type to change in the future, or why not simple use of Sender type?

crates/command-service/src/communication/config.rs
```diff
@@ -21,21 +18,27 @@ pub enum MessageQueueConfig {
...
 
-impl MessageQueueConfig {
+impl CommunicationConfig {
     pub fn task_limit(&self) -> usize {
         match self {
-            MessageQueueConfig::Kafka { task_limit, .. } => *task_limit,
-            MessageQueueConfig::Amqp { task_limit, .. } => *task_limit,
+            CommunicationConfig::Kafka { task_limit, .. } => *task_limit,
+            CommunicationConfig::Amqp { task_limit, .. } => *task_limit,
+            CommunicationConfig::Grpc { task_limit, .. } => *task_limit,
```
Kononnable, line 33: Not sure if task limiter is required with req/resp.

crates/command-service/src/input/service/message_queue.rs
```diff
@@ -1,75 +1,65 @@
...
-            consumer: consumers,
-            message_router,
-            task_limiter: TaskLimiter::new(config.task_limit()),
-            task_queue: Arc::new(ParallelTaskQueue::default()),
-        })
     }
+}
+
+#[async_trait]
+impl<P> ParConsumerHandler for MessageQueueHandler<P>
```
Kononnable, line 38: What stands `Par` for?

crates/command-service/src/main.rs
```diff
@@ -73,24 +70,11 @@ async fn start_services(
...
-                .add_service(CommandServiceServer::new(input))
-                .serve(addr.into())
-                .await?;
-        }
-    }
+    debug!("Starting command service on a message-queue");
+    MessageQueueInput::new(communication_config, message_router)
+        .await?
+        .listen()
+        .await?;
```
Kononnable, line 77: Isn't CS using grpc started in the same place?

crates/data-router/src/main.rs
```diff
@@ -73,75 +74,149 @@ async fn main() -> anyhow::Result<()> {
...
+        )
+        .await?;
 
     tokio::time::delay_for(tokio::time::Duration::from_secs(3)).await;
 
     Ok(())
 }
 
+struct Handler {
+    cache: Arc<Mutex<LruCache<Uuid, String>>>,
```
Kononnable, line 105: Weird, I thought clippy would complain about such types - 3 level of nested generics. Maybe it's fine because type names are short

crates/schema-registry/src/replication/master.rs
```diff
@@ -8,14 +8,15 @@ use tokio::{runtime::Handle, sync::oneshot};
...
     recv: mpsc::Receiver<ReplicationEvent>,
     tokio_runtime: Handle,
     mut kill_signal: oneshot::Receiver<()>,
 ) {
     let producer = match &config.queue {
-        MessageQueue::Kafka(kafka) => CommonPublisher::new_kafka(&kafka.brokers).await,
-        MessageQueue::Amqp(amqp) => CommonPublisher::new_amqp(&amqp.connection_string).await,
+        CommunicationMethod::Kafka(kafka) => CommonPublisher::new_kafka(&kafka.brokers).await,
+        CommunicationMethod::Amqp(amqp) => CommonPublisher::new_amqp(&amqp.connection_string).await,
+        CommunicationMethod::Grpc => return,
```
Kononnable, line 19: Should it log something here, or just mark as `unreachable!` or something similar?

crates/schema-registry/src/replication/mod.rs
```diff
@@ -122,16 +125,17 @@ impl ReplicationState {
...
     pub topic_or_queue: String,
     pub topic_or_exchange: String,
 }
 
 #[derive(Clone, Debug)]
-pub enum MessageQueue {
+pub enum CommunicationMethod {
     Kafka(KafkaConfig),
     Amqp(AmqpConfig),
+    Grpc,
```
Kononnable, line 138: I think adding another variant if it's still not supported is not the  best idea.

crates/utils/src/messaging_system/consumer.rs
```diff
@@ -1,20 +1,56 @@
+use crate::task_limiter::TaskLimiter;
```
Kononnable, line 1: This is no longer messaging_system, so we should probably rename whole dir or make it in two separate modules.

crates/utils/src/messaging_system/consumer.rs
```diff
@@ -99,36 +142,133 @@ impl CommonConsumer {
...
-            match self {
-                CommonConsumer::Kafka { consumer, ack_queue} => {
-                    let mut message_stream = consumer.start();
-                    while let Some(message) = message_stream.next().await {
-                        let message = message?;
-                        ack_queue.add(&message);
-                        yield Box::new(KafkaCommunicationMessage{message,consumer:consumer.clone(),ack_queue:ack_queue.clone()}) as Box<dyn CommunicationMessage>;
+    /// Process messages in order. Cannot be used with Grpc.
+    /// # Error handling
+    /// Program exits on first unhandled message. I may cause crash-loop.
```
Kononnable, line 147: I think we should return error instead of aborting  - keep aborts in main crate, not in its dependencies. It's crate job to determine how e.g. malformed message should be handled - abort/discard etc.

crates/utils/src/messaging_system/consumer.rs
```diff
@@ -99,36 +142,133 @@ impl CommonConsumer {
...
+                            log::error!("Couldn't process message: {:?}", e);
+                            std::process::abort();
+                        }
                     }
                 }
             }
+            CommonConsumer::Grpc { .. } => {
+                // Use par_run instead
+                return Err(Error::GrpcNotSupported);
+            }
```
Kononnable, line 188: Anti-pattern

crates/utils/src/messaging_system/publisher.rs
```diff
@@ -68,6 +83,28 @@ impl CommonPublisher {
                     .await?;
                 Ok(())
             }
+            CommonPublisher::Rest { url, client } => {
+                let url = url.join(&format!("{}/{}", topic_or_exchange, key)).unwrap();
```
Kononnable, line 87: Some variable names needs refactor

crates/command-service/src/input/service/message_queue.rs
frondeus: Parallel

crates/data-router/src/main.rs
frondeus: Hey, i didn't introduce this type. But I don't see any reason to change it cause as you noticed - it's short and quite readable.

crates/schema-registry/src/replication/master.rs
frondeus: We can log, true. I dont think unreachable is a way.

crates/schema-registry/src/replication/mod.rs
frondeus: And it won't be supported but we need to somehow disable replication for it

crates/utils/src/messaging_system/consumer.rs
frondeus: How would you suggest doing it then?
Introducing another type like `ParConsumer` maybe? and keeping only `Kafka`/`Amqp` variants in standard `Consumer`?

crates/utils/src/messaging_system/publisher.rs
frondeus: Whats exactly wrong with `url`?

Kononnable: I meant `topic_or_exchange`, `key` :smile:

crates/command-service/src/input/service/message_queue.rs
Kononnable: We probably shouldn't use abbreviations

crates/data-router/src/main.rs
Kononnable: Yeah, that was more like note to self then something that needs to be changed. Sorry.

crates/schema-registry/src/replication/mod.rs
Kononnable: Can we 'catch' this situation earlier? I mean throw error or just not execute code if grpc is used(validate on crate initialization) and keep replication related code without any mention of grpc

Kononnable: I mean enum which we read from input(with grpc) doesn't have to be the same enum as used in replication,

crates/utils/src/messaging_system/consumer.rs
Kononnable: Have no nice solution :| Anyway, why it can't be used with grpc? Is it because we can't limit number of messages(we're always accepting multiple connections)?

crates/api/src/events.rs
frondeus: It was quite useful for refactor, I guess I can remove it now.

crates/command-service/src/input/service/message_queue.rs
frondeus: I was inspired by par_iter in rayon but I guess we can use full `ParallelConsumerHandler` name here

crates/utils/src/messaging_system/consumer.rs
frondeus: Its because GRPC is parallel by design and can be executed on different threads, therefore you cant pass `&mut self` you need `Arc<Mutex<>>` everywhere, and that's why I introduced `ParConsumerHandler` which has `&self` requirement plus `Send + Sync + 'static`.

crates/utils/src/messaging_system/publisher.rs
frondeus: Oh so you mean `topic_or_exchange` -> `destination` or something?
Yes I agree, also in schema registry I guess we don't have `topic` anymore but more like a Topic/Queue/Address - destination maybe?

crates/utils/src/messaging_system/consumer.rs
frondeus: maybe `communication`?

crates/utils/src/messaging_system/consumer.rs
frondeus: in ordered loop sure i can return an error, but how about parallel? Is it okay to have there "nah, lets continue with next msg"?

Kononnable (dismissed):

crates/api/src/schema/context.rs
```diff
@@ -67,21 +67,21 @@ impl Context {
         }
     }
 
-    pub async fn connect_to_data_router(&self) -> anyhow::Result<CommonPublisher> {
+    pub async fn connect_to_ingestion_sink(&self) -> anyhow::Result<CommonPublisher> {
```
Kononnable, line 70: As I understand data router is used as a component, so probably better to use its name then an abstraction(Ingestion sink)

frondeus: Imho we shouldn't imply Data Router here. It can be data router, it can be however any other service that has the same interface (kafka/amqp_queue/grpc generic proto). I can see it happen in the future. The protocol used here is pretty generic.

In case of message queues we arent connecting to data router but to the kafka topic / amqp exchange.

Kononnable: I would propose:
If its general purpose - connect_to_service or something similar.
If its for cdl input - connect_to_cdl_input or something like that - if we don't want to use data router name(new component may show up in the future).

What I don't like about `ingestion_sink` here is that it doesn't really tell us anything. It tells us that we want to connect to something but not to what. For me ingestion sink isn't defined in business terms, only in technical one, and we should name methods so they reflect as much business as possible.

Kononnable (approved):

PR #284, opened 2021-03-02
ci: use local mdbook-linkcheck install script

luke-biel:


deployment/scripts/install-executable-from-git.sh
```diff
@@ -0,0 +1,187 @@
+#!/bin/sh
```
Kononnable, line 1: Is this file in any way related to CDL, or just to CI scripts?

Kononnable (commented):
I think `ci` is in most cases part of `chore`, but since we don't have that specified in any document I guess that's not a problem(readable anyway).

deployment/scripts/install-executable-from-git.sh
luke-biel: I guess to CI only

frondeus (approved):

PR #294, opened 2021-03-10
feat: edge-registry

luke-biel:
Once this PR passes I'm gonna convert all 3 TODO's into issues and write documentation.


crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[derive(Clone, Debug, StructOpt)]
+pub struct RegistryConfig {
+    #[structopt(long, env)]
+    username: String,
+    #[structopt(long, env)]
+    password: String,
+    #[structopt(long, env)]
+    host: String,
+    #[structopt(long, env, default = "5432")]
```
smores56, line 20: No need to give a default here, it's inconsistent.

crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[tonic::async_trait]
+impl EdgeRegistry for EdgeRegistryImpl {
+    async fn add_edges(
+        &self,
+        request: Request<NewEdgesMessage>,
+    ) -> Result<Response<Empty>, Status> {
+        let request = request.into_inner();
+        let conn = self.pool.get().await.unwrap();
```
smores56, line 63: Unwrap?

smores56 (changes requested):

crates/edge-registry/src/lib.rs
Kononnable: It's a draft, not ready for review yet

crates/edge-registry/src/lib.rs
luke-biel: While other variables are 90% of time deployment dependent, port rarely changes. It's no cost convinience.

crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        request: Request<SchemaId>,
+    ) -> Result<Response<RelationList>, Status> {
+        let request = request.into_inner();
+        let conn = self.connect().await?;
+
+        let schema_id = parse_as_uuid(&request.schema_id)?;
+
+        let rows = conn
+            .query(
+                "SELECT id, child_schema_id FROM relations WHERE parent_schema_id = ($1::uuid)",
```
frondeus, line 159: shouldn't we do `OR child_schema_id = ($1::uuid)`?

crates/edge-registry/src/main.rs
```diff
@@ -0,0 +1,15 @@
+use edge_registry::{EdgeRegistryImpl, RegistryConfig};
+use rpc::edge_registry::edge_registry_server::EdgeRegistryServer;
+use structopt::StructOpt;
+use tonic::transport::Server;
+
+#[tokio::main]
+async fn main() {
```
frondeus, line 7: add `env_logger::init()`

crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+};
+use std::str::FromStr;
+use std::time;
+use structopt::StructOpt;
+use tonic::{Request, Response, Status};
+use uuid::Uuid;
+
+#[derive(Clone, Debug, StructOpt)]
+pub struct RegistryConfig {
+    #[structopt(long, env)]
```
frondeus, line 17: Please add documentation comments for `--help`

crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+        let parent_schema_id = parse_as_uuid(&request.parent_schema_id)?;
+        let child_schema_id = parse_as_uuid(&request.child_schema_id)?;
+
+        let row = conn
+            .query(
+                "INSERT INTO relations (parent_schema_id, child_schema_id) VALUES ($1::uuid, $2::uuid) RETURNING id",
+                 &[&parent_schema_id, &child_schema_id]
+            )
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
```
Kononnable, line 111: Should not return unknowns to users. They might contain some sensitive information(like host address/db name)

Kononnable: Applies to similar lines bellow.

crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+                        relation_id: relation_id.to_string(),
+                        parent_schema_id: request.schema_id.clone(),
+                        child_schema_id: child_schema_id.to_string(),
+                    }
+                })
+                .collect(),
+        }))
+    }
+
+    // TODO: pagination
```
Kononnable, line 181: For now I don't think pagination is required. We can add it when there is a need for that(currently not having it is not an issue).

crates/edge-registry/src/main.rs
Kononnable: and serve metrics

crates/edge-registry/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+            let parent_object_id = parse_as_uuid(&relation.parent_object_id)?;
+            let child_object_id = parse_as_uuid(&relation.child_object_id)?;
+
+            conn
+                .query(
+                    "INSERT INTO edges (relation_id, parent_object_id, child_object_id) VALUES ($1, $2, $3)",
+                    &[&relation_id, &parent_object_id, &child_object_id],
+                )
+                .await
+                .map_err(|err| Status::internal(err.to_string()))?;
```
Kononnable, line 228: I think we should return some custom error if someone tries to add a relation between two objects if relation is not defined.

crates/edge-registry/src/main.rs
```diff
@@ -0,0 +1,15 @@
...
+    let config = RegistryConfig::from_args();
+    let registry = EdgeRegistryImpl::new(&config).await.unwrap();
+    Server::builder()
```
Kononnable, line 10: We need to be able to ingest data from MQ.

crates/edge-registry/src/lib.rs
frondeus: I saw "and write documentation." in first comment so I guess it's in backlog ;)

crates/edge-registry/src/lib.rs
luke-biel: I believe these edges are directional

crates/edge-registry/src/lib.rs
luke-biel: I will once this app is accepted. Within this PR. Just after we resolve all issues.

crates/edge-registry/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "edge-registry"
```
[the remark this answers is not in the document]
luke-biel: Registry - it's not repository. Within this PR only this name is used.

crates/edge-registry/src/main.rs
luke-biel: Added MQ listening for new edges

crates/edge-registry/src/lib.rs
luke-biel: There are 3 options:
* require users to send schema_ids along object_ids;
* query for schema in which object_id belongs;
* assuming user is responsible for keeping edge-registry intact;

I would prefer 3rd option. We can always add `safe` route later on.

(by user I mean the system, whoever will be ever responsible for sending edge data)

crates/edge-registry/src/lib.rs
luke-biel: I where they were needed.

crates/edge-registry/src/main.rs
luke-biel: done

crates/edge-registry/src/lib.rs
esavier: Triplet (relation_id schema_id schema_id) can be stored in cache later so querying is okay, but now we do not need it. I am also okay with 3'rd option. We can verify that relation_id exists.
Sending schema_ids alongside object_id is not the best idea imo.

esavier (dismissed):

deployment/compose/docker-compose.cdl.kafka.base.yml
```diff
@@ -66,6 +66,27 @@ services:
...
+      - "50110:50110"
+    environment:
+      POSTGRES_USERNAME: postgres
+      POSTGRES_PASSWORD: 1234
+      POSTGRES_HOST: postgres
+      POSTGRES_PORT: 5432
+      POSTGRES_DBNAME: postgres
+      POSTGRES_SCHEMA: cdl
+      COMMUNICATION_PORT: 50110
+      RUST_LOG: info,edge_registry=trace
```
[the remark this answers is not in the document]
luke-biel: Agreed, however this is compose. I tend to use trace here on daily basis, and in terms of tests which run on d-c right now, it's valuable.

Helmfiles are debug or info (don't remember which) so in real deployments nobody will turn trace on unless they explicitly decide to do so.

deployment/helm/cdl/values-local.yaml
```diff
@@ -25,6 +25,15 @@ schemaRegistry:
   storageSize: 3Gi
   port: 30150
 
+edgeRegistry:
```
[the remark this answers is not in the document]
luke-biel: Will do, but first thing first I want this PR accepted. Then I'll adjust documentation.

It's just that I prefer to have final view and no more code changes, so I don't have to change docs along.

crates/edge-registry/Cargo.toml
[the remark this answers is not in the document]
Kononnable: Yeas, but so far we used only `repository` when we were talking about this component.
I'm fine with using repository or with using registry, but not with both. If we want to change this name it should also affect other places(e.g. feature list) so people will at least know that we're talking about the same component.

crates/edge-registry/src/lib.rs
Kononnable: I meant only checking if relation under `relation_id` binds `schema_1` and `schema_2`.
As discussed previously we should generally store object_ids with schema_ids  since object_id is not guaranteed to be unique in different repositories.

crates/edge-registry/Cargo.toml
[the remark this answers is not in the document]
luke-biel: Agreed, updated that in rfcs

crates/edge-registry/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+        match consumer.run(handler).await {
+            Ok(_) => {
+                error!("MQ consumer finished work"); // If this happens it means that there's problem with Kafka or AMQP connection
+            }
+            Err(err) => {
+                error!("MQ consumer returned with error: {:?}", err);
+            }
+        }
+        process::abort();
+    });
```
Kononnable, line 37: can `consumer.run` panic(now or in the future)?  If so such panic won't be cought.

Kononnable (commented):
I'm not sure about edges being directional(when it comes to querying, not storage). They're for sure directional when it come to storage(otherwise we can't do any tree like structure), but as for 'standard' fetch operation it's hard to tell right now. Anyway, it will be easy to 'change' that later on in the future.

One problem with possible panic and I think we're good here.

crates/edge-registry/src/main.rs
luke-biel: I've registered panic hook in every application. We use `select!` in some places, but `metrics` for example are always run separately. This will make every panic critical.

PR #312, opened 2021-03-12
chore(workflow): automatically update branches

### LINKED ISSUE #311: chore: autoupdate branches
Automatically update pull requests

esavier:
fixes #311

### DIFF
--- a/.github/workflows/pr-auto-update.yaml
+++ b/.github/workflows/pr-auto-update.yaml
@@ -1,0 +1,17 @@
+name: autoupdate
+on:
+  push: {}
+  name: autoupdate
+jobs:
+  autoupdate:
+    name: autoupdate
+    runs-on: ubuntu-lastest
+    steps:
+      - uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          DRY_RUN: "false"
+          MERGE_MSG: "Branch was auto-updated."
+          RETRY_COUNT: "5"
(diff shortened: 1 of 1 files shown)


frondeus:
> Branch updates events caused by this action will not trigger any subsequent workflows
from docs.

That's a bit sad because if workflow merges automatically branch with no conflicts we won't know if changes present in develop wouldn't break this branch.

Kononnable:
What is the reasoning behind it? As @frondeus said it looks like it cheats CI(so no 100% safe). If another solution runs those actions its no good either - merging a PR will call CI on all other open PRs.
I just wonder why do we need that.

esavier:
> What is the reasoning behind it? As @frondeus said it looks like it cheats CI(so no 100% safe). If another solution runs those actions its no good either - merging a PR will call CI on all other open PRs.
> I just wonder why do we need that.

I would like for various reasons to have branches updated automatically not manually (i.e. always mergable or if not mergable create a conflict)

It doesn't cheat CI, problem is that workflows can trigger workflows, its a GH limitation, ther are possible workarounds though, i am working on checking those. It would be a separate trigger to restart all open PR's workflows by issuing an empty commit

Kononnable:
>If another solution runs those actions its no good either - merging a PR will call CI on all other open PRs.

I don't think we really want to start a few CI tasks for each merge to develop. Currently(9 open PRs) that would be like 30 long running(5+ minutes)  CI tasks and we can use only one PR either way(once it gets merged other tasks will be restarted anyway).

esavier:
postponed for now

PR #315, opened 2021-03-15
docs: Update changelog

frondeus:


Kononnable:
I don't think we should include following sections in our changelog (don't see any client gains from reading through those):
- Dependencies
- Internal (internal refactors)
- Tests
- Documentation - major things like github pages are worth mentioning, but not so sure about each change there
- CI and infrastructure  - same as above, internal changes are not something that client will be interested in(change in helm files sure,bumping cache action or simplifying dockerfile not so much )

esavier (approved):

esavier:
> 
> 
> I don't think we should include following sections in our changelog (don't see any client gains from reading through those):
> 
>     * Dependencies
> 
>     * Internal (internal refactors)
> 
>     * Tests
> 
>     * Documentation - major things like github pages are worth mentioning, but not so sure about each change there
> 
>     * CI and infrastructure  - same as above, internal changes are not something that client will be interested in(change in helm files sure,bumping cache action or simplifying dockerfile not so much )

* dependencies - i would like to know if a dependency bug from previous version is fixed in current
* Internal - ehhh, probably, but those two things mentioned are not fit into anywhere else, but also good to know for the end user
* Documentation - sure should be there
* ci and infra - infra is in this case example deployments. Those should not be a part of the cdl at all, but since you insist on that, then changes to those examples should be in the changelog. About stuff related to docker - client maybe wont be interested but developers relying on the project will.

PR #316, opened 2021-03-15
chore: arrange github workflows by scope, add descriptive names

luke-biel:
When this get's approved we have to change `required` workflow names


.github/workflows/test.yml
```diff
@@ -6,10 +6,108 @@ on:
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+          architecture: 'x64'
+
+      - name: Cache cargo target dir
+        uses: actions/cache@v2.1.4
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
```
Kononnable, line 97:
```suggestion
        with:
          path: target
           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
          restore-keys: ${{ runner.os }}-cargo-build-target-
```
Should work a bit better for build which changes single dependency.

Kononnable: And I agree, we should define an action for caching rust - we've got it used in few places and they're not synced.
But it's fine if this action is not be in scope for this PR

.github/workflows/test.yml
```diff
@@ -6,10 +6,108 @@ on:
       - "develop"
   pull_request:
 
+env:
```
Kononnable, line 9: We could probably limit those actions to not be started on rfc PRs/merges(ignore docs path or include just crates folder with `Cargo*` files.

frondeus:
Besides what Paweł mentioned - LGTM

esavier (dismissed):

.github/workflows/test.yml
```diff
@@ -6,10 +6,109 @@ on:
...
+        with:
+          python-version: '3.x'
+          architecture: 'x64'
+
+      - name: Cache cargo target dir
+        uses: actions/cache@v2.1.4
+        with:
+          path: target
+           key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-build-target-
```
Kononnable, line 98: Not sure how that space got here. I think it was fine in proposed suggestion
```suggestion
          path: target
          key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
          restore-keys: ${{ runner.os }}-cargo-build-target-
```

Kononnable: Ok, I see it there too, sorry :)

.github/workflows/test.yml
luke-biel: We have 2 `mandatory` actions, so I cannot (don't want to) remove running these actions on PRs, but I can omit running them on merges.

luke-biel:
https://github.com/luke-biel/CommonDataLayer/pull/4 - to compare

frondeus (approved):

PR #318, opened 2021-03-17
Formalize commit names and tags

luke-biel:
Moved from discussions


docs/rfc/CDLF-00014-00-rfc-01.md
```diff
@@ -0,0 +1,81 @@
...
+
+Tags that are used so far:
+`test`
+`ci`
+`chore`
+`refactor`
+`fix`
+`feature`
+`docs`
+`RFC`
```
Kononnable, line 30:
```suggestion
`rfc`
```

docs/rfc/CDLF-00014-00-rfc-01.md
```diff
@@ -0,0 +1,81 @@
...
+
+Some of these overlap, we foregone using the scope of the commit.
+
+## What we should be doing:
+Document the process. Create a set of `tags` that are in use.
+Some tags are duplicates. Thus I propose to use only:
+
+* `chore` - dependency updates, ci updates, refactorings, other changes that don't affect `CDL` functionality
```
Kononnable, line 38: I completely missed this last time. I don't think refactor is a chore.
Anyway I think we can use this simplified version.

docs/rfc/CDLF-00014-00-rfc-01.md
```diff
@@ -0,0 +1,81 @@
...
+
+In the meantime, we started enforcing names on PR's.
+I propose that we drop tags in PR titles altogether and use labels.
+
+As for the long commit message:
+
+for dependabot - we can leave them as is.
+For our work - we should make sure that commits titled wip aren't included; thus, it may be a good idea to have some standard there.
+
+I propose we use the format as follows:
```
Kononnable, line 55: Didn't realized it sooner. Setting up commit message is done by person who is merging(may not be the author). Are we ok with this being done by someone who is not as familiar with those changes as PR author(in our PR branches we don't have to use tags in commit messages)?

Kononnable (approved):

docs/rfc/CDLF-00014-00-rfc-01.md
luke-biel: Less is better in that case. If we ever feel that this small list is not enough we can add new tags.

docs/rfc/CDLF-00014-00-rfc-01.md
luke-biel: I could defend it by saying that reviewer should be familiar with topic, but reality verifies itself.

In complex cases it's best to delegate merging to reviewer. In simpler, it's whoever first. I don't really feel the need to create ruleset for that.

PR #320, opened 2021-03-17
chore: Replace logging with tokio::tracing

frondeus:


crates/api/src/config.rs
```diff
@@ -18,7 +18,7 @@ pub struct Config {
 
     /// Kafka topic/AMQP queue on which API listens for notifications
     #[structopt(long, env)]
-    pub report_source: String,
+    pub report_source: Option<String>,
```
frondeus, line 21: Some changes (like this one) I made to fix CDL and be able to run the whole stack locally.

crates/api/src/schema/query.rs
```diff
@@ -43,80 +44,95 @@ impl Schema {
...
-                    name: view.name,
-                    expression: view.jmespath,
+        let span = tracing::trace_span!("query_views", ?self.id);
+        async move {
+            let mut conn = context.connect_to_registry().await?;
+            let id = self.id.to_string();
+            let rpc_id = rpc::schema_registry::Id { id: id.clone() };
+
+            let views = conn
+                .get_all_views_of_schema(rpc_id.clone())
```
luke-biel, line 117:
```suggestion
                .get_all_views_of_schema(rpc_id)
```

crates/api/src/schema/query.rs
```diff
@@ -191,46 +226,56 @@ impl Query {
...
-            .collect::<Vec<CdlObject>>())
+        let span = tracing::trace_span!("query_objects", ?object_ids, ?schema_id);
+        async move {
+            let client = reqwest::Client::new();
+
+            let id_list = object_ids
+                .into_iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<String>>()
+                .join(",");
```
luke-biel, line 237: By relying on fmt::Display you can avoid allocation (this requires Itertools)

```
            let id_list = object_ids
                .iter()
                .join(",");
```

deployment/compose/docker-compose.cdl.grpc.postgres.yml
```diff
@@ -24,7 +24,8 @@ services:
       POSTGRES_PORT: 5432
       POSTGRES_DBNAME: postgres
       POSTGRES_SCHEMA: cdl
-      REPORT_DESTINATION: "http://api:50104"
+      REPORT_ENDPOINT_URL: "http://api:50104"
```
luke-biel, line 27: Shouldn't this reflect the change to api port?

crates/command-service/src/args.rs
```diff
@@ -51,7 +51,7 @@ pub struct CommunicationArgs {
     #[structopt(long, env)]
     pub amqp_consumer_tag: Option<String>,
     /// Port to listen on
-    #[structopt(long = "rpc-input-port", env = "RPC_PORT")]
+    #[structopt(long = "rpc-input-port", env = "GRPC_PORT")]
```
luke-biel, line 54: Just to be sure - this fixes the previous typo and is already reflected in docs/deployments?

luke-biel (changes requested):
I'm aware that clones were already there, but since the API port needs changing, could you fix them?

Also - I believe `edge-registry` is in and should be included. I can do that if you are busy, just give me a notice.

crates/command-service/src/args.rs
frondeus: Exactly.

deployment/compose/docker-compose.cdl.grpc.postgres.yml
frondeus: It should in theory, but this is more a placeholder - API right now supports report notifications only from Kafka or AMQP.

frondeus:
> I'm aware that clones were already there, but since the API port needs changing, could you fix them?
> 
> Also - I believe `edge-registry` is in and should be included. I can do that if you are busy, just give me a notice.

Done & done

luke-biel (approved):

Kononnable:
- How does our logging currently look like? I mean we all knew old format (`path: message`), what is displayed after those changes?
- Do we have to set `RUST_LOG` like variable to enable logging or does it support this one?
- Does our tracing reporter support also `log` syntax?  Can we enable logs from rdkafka by changing environmental variables or do we have to add back `env_logger`?

frondeus:
> * How does our logging currently look like? I mean we all knew old format (`path: message`), what is displayed after those changes?

An example

So when you use `tracing::instrument`, the message format looks like:
(`call site with arguments: path: message`) where call site can look like `Server::run{addr=0.0.0.0:50106}`.
Please keep in mind that these are structural traces, and while it might look inconvenient in the console output, we could use for example [OpenTelemetry](https://github.com/tokio-rs/tracing/tree/master/tracing-opentelemetry) to collect traces, or send them to for example GreyLog and filter them.

> * Do we have to set `RUST_LOG` like variable to enable logging or does it support this one?

You use `RUST_LOG` just like with `log` crate.

> * Does our tracing reporter support also `log` syntax?  Can we enable logs from rdkafka by changing environmental variables or do we have to add back `env_logger`?

While I didn't enable it we could use a compatibility layer to read `log::debug` from dependencies and treat them as traces.

frondeus:
> While I didn't enable it we could use a compatibility layer to read `log::debug` from dependencies and treat them as traces.

One note - it is true what I wrote, however now I see that `tracing_subscriber` has one optional dependency:

> tracing-log: Enables better formatting for events emitted by log macros in the fmt subscriber. On by default.

Therefore it should print also logs next to traces. It won't, however, send logs to `GELF`/OpenTracing etc.

luke-biel (approved):

PR #325, opened 2021-03-19
chore: Materialization - refactor views

frondeus:


crates/schema-registry/src/db.rs
```diff
@@ -331,16 +341,20 @@ impl<D: Datastore> SchemaDb<D> {
...
-            self.set_vertex_properties(id, &[(View::EXPRESSION, Value::String(jmespath))])?;
+        if let Some(materializer_addr) = view.materializer_addr {
+            self.set_vertex_properties(
+                id,
+                &[(View::MATERIALIZER_ADDR, Value::String(materializer_addr))],
+            )?;
+        }
+
+        if let Some(fields) = view.fields {
+            let fields = serde_json::to_value(fields).unwrap();
```
Kononnable, line 356: is it safe to unwrap?

Kononnable (commented):
Will it work ok if someone pass a valid json, but not valid view definition? Or will returned error  be cryptic to say the least?

crates/schema-registry/src/db.rs
frondeus: In my opinion it is perfectly safe:
Here we have `fields` variable which has type `HashMap<String, FieldDefinition>` where `FieldDefinition` is a simple enum.
And according to `serde_json` documentation:
> This conversion can fail if T's implementation of Serialize decides to fail, or if T contains a map with non-string keys.

So in that case:
* We are using default implementation of `Serialize` - there is no place to fail.
* Our hash map contains string keys.

esavier (approved):

frondeus:
> Will it work ok if someone pass a valid json, but not valid view definition? Or will returned error be cryptic to say the least?

I think I don't understand this question - if someone passes an invalid view definition then serde will return an error. Is this error cryptic? Well, serde was not designed to return human-readable messages in the first place.

Kononnable (approved):

PR #334, opened 2021-03-24
Switch to `async-graphql` for the API

smores56:
This library is more properly async, handles JSON better, and will let us switch to Tokio 1.0 finally.

### DIFF
--- a/crates/api/src/schema/context.rs
+++ b/crates/api/src/schema/context.rs
@@ -1,59 +1,36 @@
 use std::collections::HashMap;
 use std::sync::Arc;
 
-use anyhow::{Context as _, Result};
+use bb8::Pool;
 use rpc::schema_registry::schema_registry_client::SchemaRegistryClient;
 use rpc::tonic::transport::Channel;
 use tokio::sync::Mutex;
-use utils::communication::publisher::CommonPublisher;
 
-use crate::{
-    config::{CommunicationMethodConfig, Config},
-    events::EventStream,
-    events::EventSubscriber,
-};
+use crate::{config::Config, events::EventStream, events::EventSubscriber};
 
 #[derive(Clone)]
-pub struct Context {
-    config: Arc<Config>,
-    mq_events: Arc<Mutex<HashMap<String, EventSubscriber>>>,
+pub struct MQEvents {
+    pub events: Arc<Mutex<HashMap<String, EventSubscriber>>>,
 }
 
-impl juniper::Context for Context {}
-
-impl Context {
-    pub fn new(config: Arc<Config>) -> Self {
-        Context {
-            config,
-            mq_events: Default::default(),
-        }
-    }
-
-    pub fn config(&self) -> &Config {
-        &self.config
-    }
-
-    #[tracing::instrument(skip(self))]
-    pub async fn connect_to_registry(&self) -> Result<SchemaRegistryConn> {
-        tracing::debug!("Connecting to registry");
-        // TODO: Make proper connection pool
-        let new_conn =
-            rpc::schema_registry::connect(self.config.schema_registry_addr.clone()).await?;
-        Ok(new_conn)
-    }
-
-    pub async fn subscribe_on_communication_method(&self, topic: &str) -> Result<EventStream> {
+impl MQEvents {
+    pub async fn subscribe_on_communication_method(
+        &self,
+        topic: &str,
+        config: &Config,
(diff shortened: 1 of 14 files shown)


crates/api/src/schema/context.rs
```diff
@@ -68,24 +45,32 @@ impl Context {
...
+    type Error = rpc::error::ClientError;
+
+    async fn connect(&self) -> Result<Self::Connection, Self::Error> {
+        tracing::debug!("Connecting to registry");
+
+        rpc::schema_registry::connect(self.address.clone()).await
+    }
+
+    async fn is_valid(&self, conn: Self::Connection) -> Result<Self::Connection, Self::Error> {
+        Ok(conn)
```
frondeus, line 66: Are you sure that this is always true?

crates/api/src/schema/context.rs
```diff
@@ -68,24 +45,32 @@ impl Context {
...
+    }
+
+    fn has_broken(&self, _conn: &mut Self::Connection) -> bool {
+        false
```
frondeus, line 70: Also this one. Shouldn't we check if the connection is alive somehow?

crates/api/src/types/schema.rs
```diff
@@ -46,7 +48,7 @@ pub struct View {
 }
```
frondeus, line 48: `fields` should be `Json<Value>`

crates/command-service/Cargo.toml
```diff
@@ -25,7 +25,7 @@ bb8         = "0.4"
 fnv         = "1.0"
 futures     = "0.3"
 rdkafka     = { version = "0.24", features = ["cmake-build"] }
-reqwest     = "0.10"
+reqwest     = "=0.10.10"
```
frondeus, line 28: Is there a reason why you pinned the version? If so, could you write a comment in Cargo.toml explaining it? For example, link to an issue in Github that requires it (so in the future another programmer can see it, check, and eventually bump)

frondeus (changes requested):

crates/api/src/schema/mutation.rs
```diff
@@ -1,42 +1,34 @@
...
-pub struct Mutation;
+pub struct MutationRoot;
 
-#[graphql_object(context = Context)]
-impl Mutation {
-    async fn add_schema(context: &Context, new: NewSchema) -> FieldResult<Schema> {
+#[Object]
+impl MutationRoot {
+    async fn add_schema(&self, context: &Context<'_>, new: NewSchema) -> FieldResult<Schema> {
         let span = tracing::trace_span!("add_schema", ?new);
```
frondeus, line 19: One more thing, with new `#[Object]` macro attribute, could you check whether `tracing::instrument` attribute works properly?
It does not for `graphql_object` but I know it can for other macros like `async_trait`.
If so, we could then replace
`let span = tracing::trace_span!` with `tracing::instrument` macro attribute (just like in for example `schema-registry/src/rpc.rs`) for 
making code much clearer and prone to error.
Of course if you see it as too big change - I could introduce it later in another PR.

smores56: Yeah, I tried it and it didn't work. There might be a way to do it, but it will have to be in another PR with more time allotted.

crates/command-service/Cargo.toml
smores56: This PR is the "first half" of the `tokio` 1.0 conversion. To ensure that everything was on the same `tokio` version, I pinned the version here to crates that use 0.2. In the `tokio` 1.0 PR, this gets upgraded to the latest version.

crates/api/src/types/schema.rs
smores56: Converted all usage to `Json<Value>`.

crates/api/src/schema/context.rs
smores56: This method is called when retrieving a connection from the pool, so I sent a `heartbeat` request to ensure that the connection is valid before using it.

crates/api/src/schema/context.rs
smores56: This method is called before putting connection back into the pool. Since we are already checking whether a connection is valid when we retrieve it from the pool, and we just used it before we put it back, I am going to leave this for now rather than make 2 extra RPC requests every time we use the Schema Registry.

frondeus (approved):

PR #335, opened 2021-03-25
Update to Tokio 1.0

smores56:
With the change to `async-graphql` for our GraphQL API, we can now finally move all services to `tokio` 1.0 or later.

Most files are simply version changes, with the exception of `bb8`, `metrics`, and `indradb` (which is now finally using a version hosted on cargo instead of a git commit).

If you want to verify that this PR actually moves everything to `tokio` 1.0, you can run `rg "tokio 0" Cargo.lock"` and verify that there is no output.


crates/schema-registry/src/main.rs
```diff
@@ -173,17 +174,21 @@ pub async fn main() -> anyhow::Result<()> {
     .await?;
 
     if let Some(export_dir_path) = config.export_dir {
-        let exported = registry.export_all()?;
+        let exported = registry
+            .export_all()
+            .map_err(|err| anyhow::anyhow!("{}", err))?;
```
frondeus, line 179: What's the point of wrapping it in a string?

smores56: Our `thiserror`-defined error type includes a variant that isn't Sync, so anyhow can't automatically convert with the Context trait. So, we have to wrap manually to return `anyhow::Error`'s.

crates/schema-registry/Cargo.toml
```diff
@@ -23,7 +23,7 @@ anyhow      = "1.0"
 async-trait = "0.1"
 chrono      = "0.4"
 futures     = "0.3"
-indradb-lib = { git = "https://github.com/jespersm/indradb", branch = "sled", features = ["sled-datastore"] } # We use fork due to reported incompatibility of sled 0.34 with azure k8s. More testing needed before switching to official build.
+indradb-lib = { version = "2.1", features = ["sled-datastore"] }
```
Kononnable, line 26: Was it tested? I mean comment describes that standard version didn't work correctly with azurefiles, so was it tested there, or do we no longer care about this scenario(I'm fine with that, just need to make it more obvious).

Kononnable (dismissed):

crates/schema-registry/Cargo.toml
esavier: from my understanding we can get away with this even if its not working. We are getting rid of indra soon either way.
i am doing the manual test in the background but i will not spend too much time on that for the same reasons

esavier (commented):
seems okay, I am doing some manual testing, please stand by

crates/schema-registry/Cargo.toml
esavier: okay, it seems to work, SR will be superseeded with PGxx based one so it should be okay.

esavier (approved):
everything seems to be okay

PR #337, opened 2021-03-26
feat: Object builder

### LINKED ISSUE #330: Implement trivial object builder, no filtering at this point, return objects for each object_id in view_id base schema
```
Input:
{
view_id: UUID,
}
Output: 
{
view_id: UUID,
rows:RowDefinition
}
RowDefinition - array of objects where each object returns 
(object_id, [field:value])
```
Input - both pub/sub and req/resp, for output req/resp(may be initialized by ObjectBuilder)

frondeus:
Closes #330 
It does not send data anywhere as we lack an interface - I guess it would be via GRPC.
However I tested it manually, pushed some data, and checked logs - it builds materialized view nicely.


crates/object-builder/src/service.rs
```diff
@@ -0,0 +1,193 @@
...
+        Ok(Self {
+            consumer,
+            handler: ServiceHandler { schema_registry },
+        })
+    }
+
+    pub async fn listen(self) -> anyhow::Result<()> {
+        self.consumer.par_run(self.handler).await?;
+
+        tokio::time::delay_for(tokio::time::Duration::from_secs(3)).await;
```
Kononnable, line 189: It's not related to service itself, it's related to whole program lifecycle, so it should go to main.

crates/utils/src/communication/parallel_consumer.rs
```diff
@@ -104,16 +104,24 @@ impl ParallelCommonConsumer {
                 )
                 .await
             }
-            ParallelCommonConsumerConfig::Grpc { addr } => Ok(Self::Grpc { addr }),
+            ParallelCommonConsumerConfig::Grpc { addr } => Self::new_grpc(addr),
         }
     }
 
+    #[tracing::instrument]
+    fn new_grpc(addr: SocketAddrV4) -> Result<Self> {
```
Kononnable, line 112: Method is not failing so should not return a result.

deployment/compose/setup/schema_registry/initial-schema.json
```diff
@@ -39,13 +44,13 @@
...
             "query_address": "http://postgres_query:50201",
-            "insert_destination": "cdl.document.data",
+            "insert_destination": "http://postgres_command:50202",
             "name": "new schema"
         },
         "a5e0c7e2-412c-11eb-8000-000000000000": {
             "schema_type": "DocumentStorage",
             "query_address": "http://postgres_query:50201",
-            "insert_destination": "cdl.document.data",
+            "insert_destination": "http://postgres_command:50202",
```
Kononnable, line 53: Necessary changes or just leftovers from testing?

Kononnable (changes requested):
Had to check PR title multiple times - first 30% of diff shown by github is refactor completely unrelated to PR. One change really, but it takes a lot of space. Just mentioning, no need to change that now.

- No k8s definitions

crates/utils/src/lib.rs
```diff
@@ -16,9 +16,14 @@ pub mod psql;
...
-        tracing_subscriber::fmt::init();
+pub mod tracing;
+pub mod types {
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+    #[serde(rename_all = "snake_case")]
+    pub enum FieldDefinition {
+        FieldName(String),
     }
```
Kononnable, line 27: Not sure if that should go to the `lib.rs` or to some other module(or do we need to create new one). It doesn't feels that global to me to go to `lib.rs`.

crates/object-builder/src/service.rs
```diff
@@ -0,0 +1,193 @@
...
+    view_id: Uuid,
+    rows: Vec<RowDefinition>,
+}
+
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "snake_case")]
+struct RowDefinition {
+    object_id: Uuid,
+    fields: HashMap<String, Value>,
+}
```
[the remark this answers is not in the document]
frondeus: As soon as I'll prepare first materializer then I'll move it to utils.

> why field is Value?

I don't understand what do you expect here. An enum? Even when we know data type (because we have schema with definition in hand) we still need to serialize it somehow, therefore a `Value`. It can be string, number, array, whatever user defines in its schema. I'd go and just send it to the user, and if the data is mismatched with its type in schema, then he should use validator during ingestion in my opinion.

crates/object-builder/src/args.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[structopt(long, env)]
+    pub schema_registry_addr: String,
+    /// Port to listen on for Prometheus requests
+    #[structopt(long, default_value = utils::metrics::DEFAULT_PORT, env)]
+    pub metrics_port: u16,
+}
+
+#[derive(Error, Debug)]
+#[error("Missing config variable `{0}`")]
+pub struct MissingConfigError(pub &'static str);
```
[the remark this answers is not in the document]
frondeus: well we have similar logic in command service, should we panic there as well?

crates/object-builder/src/service.rs
```diff
@@ -0,0 +1,193 @@
...
+        let rows = objects
+            .into_iter()
+            .map(|(object_id, object)| Self::build_row_def(object_id, object, &fields_defs))
+            .collect::<anyhow::Result<_>>()?;
+
+        let output = Output { view_id, rows };
+
+        let _materializer_addr = view.materializer_addr;
+        // TODO: Sending output to materializer
+        // Via GRPC?
```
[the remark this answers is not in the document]
frondeus: Ill send data to materializer as soon as I define one, if you don't mind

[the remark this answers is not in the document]
frondeus: GRPC - sending in response  - done

deployment/compose/setup/schema_registry/initial-schema.json
frondeus: moved changes to separate `initial-schema.grpc.json` while the original one became `initial-schema.mq.json`. It solves the problem and can be useful. Thanks for pointing it

crates/object-builder/src/args.rs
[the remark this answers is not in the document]
Kononnable: From my perspective non-recoverable error = panic, if it can be catched and changed to some default value such error type would be fine. Anyway, keeping code more consistent is more important then this change, so for my taste - either change both(if this error isn't being cought anywhere) or keep it as it is.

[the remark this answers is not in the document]
frondeus: I'll leave it and later create another PR cleaning it.

crates/object-builder/src/service.rs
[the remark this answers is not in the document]
Kononnable: I think we're making this more complex then it needs to be. If it's just a `string` it's fine. Problems with `value` is that we're introducing one more place we're depending on another crate and the value needs to be serialized in more complex way(`serde_json` validates json on object creation etc.).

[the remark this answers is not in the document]
frondeus: We still need to parse objects to some kind of structure to retrieve required fields.
How about using `RawValue` instead?

Kononnable (commented):
Looks fine, but I think one more person should look at this because it can get complex really quick(e.g. we had some discussions during review).

crates/object-builder/src/service.rs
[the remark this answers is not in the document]
frondeus: Okay, the main question is:

> Do we allow having a view where the field points to a complex data structure like an array or a JSON object, or only primitives like strings & numbers? 

Before we answer that we cannot assume anything about this `HashMap`.

Another thing is we have to keep in mind that this view field might be a result of some aggregation or calculation, for example `avg()`.

We could keep everything in the String, and therefore output json would be:
```
{
  "object_id": "...",
  "fields": {
      "foo": "[ 1, 2, 3 ]", //complex scenario
      "bar": "5", //primitive scenario
   }
}
```
But personally, I think it might not be the best idea from the consumer point of view.
I'd use `RawValue` or `Value` here. Then the consumer can easier distinguish if `"5"` means `5` or string with a glyph `'5'`. Which might not **but might be** important for them.

Of course, type info is in the schema so we could, in theory, have some kind of `WhatTypesReturnsThisView` query which would do basically type inference, but it requires extra work and extra call to the schema registry which we might want to avoid.

Kononnable (dismissed):

crates/rpc/proto/object_builder.proto
```diff
@@ -0,0 +1,20 @@
...
+}
+
+message MaterializedView {
+  required string view_id = 1;
+  repeated RowDefinition rows = 2;
+}
+
+message RowDefinition {
+  required string object_id = 1;
+  map<string, bytes> fields = 2;
```
frondeus, line 19: after writing tests i can see why `bytes` might not be the best representation.
However I did some research and in theory `proto3` documentation tells me to use `Any` type for any JSON object...

Which seems to be unsuported by `prost` :(

Kononnable (approved):

crates/rpc/proto/object_builder.proto
frondeus: changed to `string` for now

Cargo.lock
```diff
@@ -98,9 +98,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.40"
+version = "1.0.39"
```
luke-biel, line 101: Why it's downgraded?

crates/object-builder/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "snake_case")]
+struct Output {
+    view_id: Uuid,
+    rows: Vec<RowDefinition>,
+}
+
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "snake_case")]
```
luke-biel, line 36: Isn't this a default?

Cargo.lock
frondeus: I... don't know, it wasn't a conscious decision. I don't know why cargo had chosen `.39`. Anyway, ran `cargo update -p anyhow`

luke-biel (approved):

crates/object-builder/src/lib.rs
frondeus: No, the default for json is `camelCase`

PR #349, opened 2021-03-31
chore: update rust-toolchain to 25-03-2021

luke-biel:


luke-biel:
This is to get rid of complier panics introduced in previous toolchain bump.

frondeus (approved):

PR #357, opened 2021-04-06
partial-update-engine

### LINKED ISSUE #333: Implement Partial Update Engine for object changes(no relation support yet)
- [ ] Run periodically, expose sleep time as parameter or k8s config(what with other deployment types?)
- [ ] For each notification check if view exists which is based on it(uses same schema_id). If so add such record to the change list(separate list per view)
- [ ] After all existing notifications are read(EOF) reduce change list so it don't contain duplicates(doesn't have to be done in separate stage)
- [ ]  Send view rebuild request to object builder, ignore change list for the time being
- [ ] Enter sleep phase

Kononnable:
Closes #333

First implementation of partial update engine, code is complete, but not yet e2e tested.


crates/partial-update-engine/src/main.rs
```diff
@@ -0,0 +1,191 @@
...
+        .set("enable.partition.eof", "false")
+        .set("session.timeout.ms", "6000")
+        .set("enable.auto.commit", "true")
+        .set("enable.auto.offset.store", "false")
+        .set("auto.offset.reset", "earliest")
+        .create()
+        .context("Consumer creation failed")?;
+    let topics = [config.notification_topic.as_str()];
+
+    rdkafka::consumer::Consumer::subscribe(&consumer, &topics)
```
frondeus, line 75: Shouldn't we support RabbitMQ as well?

crates/partial-update-engine/src/main.rs
```diff
@@ -0,0 +1,191 @@
...
+        .set("acks", "all")
+        .set("compression.type", "none")
+        .set("max.in.flight.requests.per.connection", "5")
+        .create()?;
+
+    let mut message_stream = Box::pin(consumer.stream().timeout(Duration::from_secs(2))); // TODO: configure?
+    let mut changes: HashSet<PartialNotification> = HashSet::new();
+    let mut offsets: HashMap<i32, i64> = HashMap::new();
+    loop {
+        // TODO: configure max items per batch(?) - otherwise we won't start view recalculation if messages are sent more often then timeout
```
frondeus, line 90: Is this TODO for phase 2 or for now?

crates/partial-update-engine/src/main.rs
```diff
@@ -0,0 +1,191 @@
...
+            .iter()
+            .map(|view| Ok(view.0.parse()?))
+            .collect::<Result<Vec<Uuid>>>()?;
+        for view in schema_in_views {
+            views.insert(view);
+        }
+    }
+
+    for view in views {
+        let payload = format!("{{{}}}", view); // TODO: Use same type as object builder
```
frondeus, line 160: Object builder uses simplest possible `view_id: Uuid = str_from_mq.trim().parse()?`. There is no JSON format involved, just a simple UUID.

deployment/helm/cdl/values-local.yaml
```diff
@@ -43,6 +43,11 @@ objectBuilder:
 webAdmin:
   port: 30151
 
+partialUpdateEngine:
+  objectBuilderTopic: cdl.object-builder.input
```
frondeus, line 47: Either change it to `cdl.materialize`, or change line 40 to `cdl.object-builder.input`

crates/partial-update-engine/src/main.rs
Kononnable: I think for now it's not a requirement - it wasn't mentioned on any planning meetings, and using notifications(or any one-to-many connections) there is a bit more complex(in terms of performance and resource usage) for scenarios where we process messages in batches.
I think it's safe to make materialization work on kafka only and discuss enabling it on rabbit/grpc approach once feature is implemented.

crates/partial-update-engine/src/main.rs
Kononnable: For later on, not sure if it's phase 2, or when. Not yet sure how to properly address this - probably some config variable or something similar because how it should work is really affected by the workflow.

frondeus (approved):

PR #360, opened 2021-04-07
chore: general workflow improvements

luke-biel:
This is a proposition - as a result, we'd push images per `tag`, eg. `1.2.4` and with each `develop` push, by commit hash, eg. `dev-2021-04-14-abcdef01`.

We may also consider nightly builds with cron, eg. if there was change since yesterday, then let's build and push `dev-2021-04-14`. I can change to that. It's team call.


esavier:
Technically okay, 
alternatively, we can limit this to build per day, and only make a tag'n'push when something was pushed this specific day.

luke-biel:
> Technically okay,
> alternatively, we can limit this to build per day, and only make a tag'n'push when something was pushed this specific day.

Now this should work like so.

esavier:
okay i forgot about it
we have to make a filter to react only on (fix) and (feat) an ignore (chore)s

PR #369, opened 2021-04-12
Update audit-check.yml

luke-biel:
We should run audit only on pushes to develop and main

### DIFF
--- a/.github/workflows/audit-check.yml
+++ b/.github/workflows/audit-check.yml
@@ -1,6 +1,9 @@
 name: Audit
 on:
   push:
+    branches:
+      - develop
+      - main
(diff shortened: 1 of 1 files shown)


Kononnable:
Just to be sure: You propose that audit check is just informational, shouldn't 'block' introducing new vulnerable dependencies in PRs?

luke-biel:
We already run the audit on each pull request. Before this PR we also run it on each push. So in the case of PR, we run it twice. Apparently, GitHub doesn't like two audits running at the same time (?) and one of the checks fails and needs to be manually restarted.

frondeus (approved):

PR #372, opened 2021-04-12
something something dark side

esavier:
thingies

### DIFF
--- a/crates/schema-registry/migrations/20210215170655_init-db.sql
+++ b/crates/schema-registry/migrations/20210215170655_init-db.sql
@@ -1,4 +1,13 @@
-CREATE TYPE schema_type_enum AS ENUM ('document_storage', 'timeseries');
+-- yes we are dropping all
+-- if you want your trash untouched write your own migration
+DROP TABLE IF EXISTS schemas CASCADE;
+DROP TABLE IF EXISTS views CASCADE;
+DROP TABLE IF EXISTS definitions CASCADE;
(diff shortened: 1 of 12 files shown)


crates/schema-registry/migrations/20210215170655_init-db.sql
```diff
@@ -1,4 +1,13 @@
-CREATE TYPE schema_type_enum AS ENUM ('document_storage', 'timeseries');
+-- yes we are dropping all
+-- if you want your trash untouched write your own migration
```
smores56, line 2: pls

smores56 (changes requested):

crates/schema-registry/migrations/20210215170655_init-db.sql
esavier: please remove or write more? :3

smores56 (approved):

crates/schema-registry/src/main.rs
```diff
@@ -18,6 +19,7 @@ pub async fn main() -> anyhow::Result<()> {
     let config = Config::from_args();
 
     status_endpoints::serve();
+    sleep(Duration::from_millis(500)).await;
```
frondeus, line 22: Why this sleep is necessary?

esavier: it prevents the dark magic of mordor from finding Frodo

PR #373, opened 2021-04-14
feat: Open Telemetry

frondeus:


crates/object-builder/src/lib.rs
```diff
@@ -131,14 +134,15 @@ impl ConsumerHandler for ObjectBuilderImpl {
 
         rpc::materializer::connect(view.materializer_addr)
             .await?
-            .upsert_view(rpc_object)
+            .upsert_view(utils::tracing::grpc::inject_span(rpc_object))
```
Kononnable, line 137: Looks kind of weird, can we move tracing call to upsert_view method or run it as separate statement(not as nested one)?

crates/utils/src/tracing.rs
```diff
@@ -1,3 +1,25 @@
...
+pub mod grpc;
+pub mod kafka; // Used mostly in common publisher and [parallel_]consumer
+
 pub fn init() {
-    tracing_subscriber::fmt::init();
+    global::set_text_map_propagator(TraceContextPropagator::new());
+
+    let tracer = opentelemetry_jaeger::new_pipeline()
+        .install_batch(opentelemetry::runtime::Tokio)
+        .unwrap();
```
Kononnable, line 13: Not sure if unwrap is a correct behavior here. Telemetry isn't essential to any service - even if it fails services will work correctly(only with less debugging output), so we'll be fine with just reporting an error(without panic)

crates/utils/src/tracing.rs
```diff
@@ -1,3 +1,25 @@
...
+    let opentelemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+
+    let fmt = tracing_subscriber::fmt::layer();
+
+    tracing_subscriber::registry()
+        .with(tracing_subscriber::EnvFilter::from_default_env())
+        .with(fmt)
+        .with(opentelemetry)
+        .try_init()
+        .unwrap();
```
Kononnable, line 24: Same here

crates/api/src/schema/mutation.rs
```diff
@@ -16,7 +16,7 @@ pub struct MutationRoot;
 #[Object]
 impl MutationRoot {
     async fn add_schema(&self, context: &Context<'_>, new: NewSchema) -> FieldResult<Schema> {
-        let span = tracing::trace_span!("add_schema", ?new);
+        let span = tracing::info_span!("add_schema", ?new);
```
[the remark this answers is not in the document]
frondeus: It should be that way, but unfortunately `#[Object]` procedural macro is hostile towards other procedural macros, so I cannot use `tracing::instrument` :( I could however try and make an OSS contribution to make them work.

crates/object-builder/src/lib.rs
frondeus: Upsert_view method is generated by prost... I wish there was simpler solution for injecting data there.
One thing I could do is to prepare custom implementation to `Materializer` trait which would inject span and call grpc client, but then we have another structure to maintain.

crates/utils/src/tracing.rs
frondeus: This one not necessarily. I would say that having no logs/traces is a critical error. We cannot be blind.

crates/utils/src/tracing.rs
frondeus: Sure

crates/api/src/schema/mutation.rs
```diff
@@ -279,7 +279,7 @@ impl MutationRoot {
         context: &Context<'_>,
         relations: Vec<ObjectRelations>,
     ) -> FieldResult<bool> {
-        let span = tracing::trace_span!("add_edges", len = relations.len());
+        let span = tracing::info_span!("add_edges", len = relations.len());
```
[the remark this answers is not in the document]
frondeus: @luke-biel ?

Kononnable (approved):

crates/api/src/schema/mutation.rs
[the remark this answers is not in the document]
luke-biel: Order of macros matter, did you try both placing `tracing` before and after Object?

luke-biel (dismissed):

crates/api/src/schema/mutation.rs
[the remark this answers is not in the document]
frondeus: `tracing::instrument` should be placed for `add_schema` method, while `#[Object]` is for the `impl MutationRoot`.

Kononnable (approved):

PR #377, opened 2021-04-16
feat: Add resolve tree route

luke-biel:


crates/edge-registry/src/lib.rs
```diff
@@ -190,6 +194,17 @@ impl EdgeRegistryImpl {
         counter!("cdl.edge-registry.get-edge", 1);
         let conn = self.connect().await?;
 
+        self.get_edge_with_conn(&conn, relation_id, parent_object_id)
+            .await
+    }
+
+    #[tracing::instrument(skip(self))]
```
frondeus, line 201: I think we'd like to skip also `conn`

frondeus (dismissed):

crates/edge-registry/src/lib.rs
luke-biel: Done

frondeus (approved):

PR #378, opened 2021-04-16
chore: Move prost codegen from build.rs to xtask

frondeus:
Because we are not changing proto file definitions that often it makes sense to move code generation from `build.rs` into separate manual command. It gives few benefits:
* Build should be a little bit faster.
* All features of IDE should now work OOB because generated code is a part of our source tree.
  I mean code completion, go-to-definition etc.
* Technically we could further modify those files by another step to for example inject open telemetry spans automatically (instead of calling `utils::tracing::grpc::inject_span` everywhere)


frondeus:
I'd suggest skipping `codegen` module checking because it is exactly 1:1 of what prost generated, just now checked by git. Everything still builds and I tested it manually, it gives same results.

esavier (approved):

esavier:
retested manually, seems to work, approved
nice one ;)

PR #391, opened 2021-04-20
feat: Filter by object ids in object builder

frondeus:


crates/api/src/types/view.rs
```diff
@@ -21,3 +21,46 @@ pub struct RowDefinition {
     /// Materialized fields
     pub fields: HashMap<String, Json<Value>>,
 }
+
+#[derive(Debug, Clone, InputObject)]
+pub struct OnDemandViewRequest {
+    /// View's UUID
+    pub view_id: Uuid,
+    /// Schemas with objects. This collection is treated like a hash-map with `schemaId` as a key, therefore `schemaId` should be unique per request.
```
frondeus, line 29: cannot use hashmap here because graphql treats hashmap keys as `A-Za-z_`, so uuid fails miserably :|

crates/api/src/types/view.rs
```diff
@@ -21,3 +21,46 @@ pub struct RowDefinition {
...
+    pub schemas: Vec<Schema>,
+}
+
+//async_graphql::scalar!(OnDemandViewRequest);
```
frondeus, line 33:
```suggestion
```

crates/object-builder/src/lib.rs
```diff
@@ -143,26 +143,33 @@ impl ConsumerHandler for ObjectBuilderImpl {
...
 
-        tracing::debug!(?view, ?base_schema, "View");
+        let view = self.get_view(&view_id).await?;
+        tracing::debug!(?view, "View");
 
-        let options = serde_json::from_str(&view.materializer_options)?;
+        // For now view can have only one base schema,
+        // and it have to have at least one
+        // Therefore panic is safe.
+        // TODO: Handle more than one schema
```
frondeus, line 157:
```suggestion
        // TODO: Handle more than one schema
        // TODO: Handle empty filter for seeding view (maybe in another method)
```

crates/object-builder/src/lib.rs
```diff
@@ -250,16 +271,16 @@ impl ObjectBuilderImpl {
     #[tracing::instrument(skip(self))]
     async fn get_base_schema(
```
frondeus, line 272:
```suggestion
    // TODO: Change name to `get_schema_metadata`
    async fn get_base_schema(
```

crates/object-builder/src/lib.rs
```diff
@@ -143,26 +143,34 @@ impl ConsumerHandler for ObjectBuilderImpl {
...
+        let view = self.get_view(&view_id).await?;
+        tracing::debug!(?view, "View");
 
-        let options = serde_json::from_str(&view.materializer_options)?;
+        // For now view can have only one base schema,
+        // and it have to have at least one
+        // Therefore panic is safe.
+        // TODO: Handle more than one schema
+        // TODO: Handle empty filter for seeding view (maybe in another method)
+        let (schema_id, schema) = schemas.into_iter().next().unwrap();
```
frondeus, line 159: ok, panic is not "safe" - schemas are for requested schemas and not for "schemas linked to view", my bad. For now I focused on integration between partial update engine and object materialize. I ignored on-demand because.... there is no on demand materializer yet. I have another issue in-progress and then ill handle this issue.

So therefore while it can panic when someone do `empty` filter, I wouldn't change it now, especially that this `next()` would disappear when we introduce joins.

Plus I haven't decided yet whether `empty filter` would be solved in object builder or on-demand materializer.

frondeus: So I guess i just need to change misleading TODO

luke-biel: Note this in any related task on our board. TODO can be forgotten.

luke-biel (approved):

PR #401, opened 2021-04-22
feat: On demand materializer

frondeus:


crates/materializer-ondemand/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+    utils::status_endpoints::serve(args.status_port);
+    utils::metrics::serve(args.metrics_port);
+
+    let materializer = MaterializerImpl::new(&args).await?;
+
+    utils::status_endpoints::mark_as_started();
+
+    Server::builder()
+        .add_service(OnDemandMaterializerServer::new(materializer))
+        .serve(([0, 0, 0, 0], args.input_port).into())
```
esavier, line 24: would be nice later to control which interface we are available on

esavier (approved):

crates/materializer-ondemand/src/main.rs
frondeus: We can create a task later - its something we would have to handle in all services, not only in materializer.

PR #402, opened 2021-04-26
feat: edge registry notifications

luke-biel:


crates/utils/src/message_types.rs
```diff
@@ -1,9 +1,13 @@
 use serde::{Deserialize, Serialize};
 use serde_json::value::RawValue;
-use serde_json::Value;
 use uuid::Uuid;
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
+pub trait OwnMessage {
```
luke-biel, line 5: I would love to use ToOwned trait, but it requires Borrow in opposite direction :/

esavier (approved):

PR #407, opened 2021-04-27
feat: SchemaRegistry - drop import/export functionality

esavier:


esavier:
Amount of work here will depend on testing and test harness for component testing. Right now dropping this will mean rewriting a lot of the tests that may still be rewritten after changing CI. Let's par this for now until the situation will clear out.

PR #408, opened 2021-04-28
fix: Query multiple returns error for timeseries

### LINKED ISSUE #52: bug: Query service doesn't validate schema_id on QueryMultiple route
As in title. It seems as something that some day may cause problems. But may as well not. We need to discuss whether to ignore this behaviour or change it.

frondeus:
fixes #52

### DIFF
--- a/crates/query-router/src/handler.rs
+++ b/crates/query-router/src/handler.rs
@@ -80,12 +80,22 @@ pub async fn query_multiple(
     schema_id: Uuid,
     cache: Arc<SchemaRegistryCache>,
 ) -> Result<impl warp::Reply, warp::Rejection> {
-    let (query_address, _) = cache.get_schema_info(schema_id).await?;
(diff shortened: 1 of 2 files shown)


crates/query-router/src/handler.rs
```diff
@@ -80,12 +80,22 @@ pub async fn query_multiple(
...
-    let (query_address, _) = cache.get_schema_info(schema_id).await?;
+    let (query_address, schema_type) = cache.get_schema_info(schema_id).await?;
 
     let object_ids = object_ids.split(',').map(str::to_owned).collect();
-    let values = rpc::query_service::query_multiple(object_ids, query_address.clone())
-        .await
-        .map_err(Error::ClientError)?;
+
+    let values = match schema_type {
+        SchemaType::DocumentStorage => {
```
Kononnable, line 88: `If let` might be nicer if only DS is allowed here.

Kononnable (approved):

PR #409, opened 2021-04-28
chore: Add opentelemetry for HTTP warp servers

frondeus:
* Remove bug: std::process:abort()
* Add tokio::time::sleep on exit
Waits for #408


crates/utils/Cargo.toml
```diff
@@ -16,7 +16,7 @@ rpc         = { path    = "../rpc" }
 # Crates.io
```
luke-biel, line 16: formatting

luke-biel (changes requested):

luke-biel (approved):

PR #410, opened 2021-04-28
chore: Easier to use grpc telemetry 

frondeus:


crates/rpc/src/materializer_general.rs
```diff
@@ -5,10 +5,21 @@ use tonic::transport::Channel;
...
-    GeneralMaterializerClient::connect(addr)
+    connect_inner(addr)
         .await
         .map_err(|err| ClientError::ConnectionError {
             service: "materializer_general",
             source: err,
         })
 }
+
+async fn connect_inner(
```
Kononnable, line 16: Not sure if its worth to extract such functions just for the purpose of having one-liners.

crates/tracing_tools/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+version = "0.1.0"
+edition = "2018"
+license = "Apache-2.0"
+
+[lib]
+name = "tracing_tools"
+path = "src/lib.rs"
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
```
Kononnable, line 13: Are feature flags really needed here? I think we're always using all of those for every affected crate. Even if not we're using all of those for the whole workspace, so crate will always be compiled with every available feature and not required functions will be removed later on.

Kononnable (approved):

crates/rpc/src/materializer_general.rs
frondeus: I extracted it to a separate function so I don't have to `map_err` every single time.
TBH, I thought about making these `connect()` in `cargo xtask codegen` but I don't have enough time :)

crates/tracing_tools/Cargo.toml
frondeus: Yeah, for now it makes little sense, but later i'd love to split utils into smaller crates so `data-router` wouldn't use `grpc` feature at all.

PR #413, opened 2021-04-29
feat: load configs from files

luke-biel:


luke-biel:
superseeded by https://github.com/epiphany-platform/CommonDataLayer/pull/470

PR #450, opened 2021-05-05
fix: PartialNotification format in PUE

### LINKED ISSUE #444: bug: PUE - invalid message format

frondeus:
fixes #444


luke-biel (approved):
We should publish & share notification format at some point.

PR #455, opened 2021-05-06
chore: remove vec_map

Kononnable:
vecmap allocates way too much memory if index value is only increasing


smores56 (approved):

luke-biel (approved):
I can let it through, but we are swapping to other Map for memory usage optimization (which I assume was tested manually, we don't have automatic metrics) with cost in performance (also, no metrics)

PR #457, opened 2021-05-07
feat: Use config crate as base for cdl configuration, in place of clap and cli

luke-biel:
Most of these changes are refactorings. I've unified way we manage our configs. No more cli - it started to become messy, in some places we had need for multiple subcommands.

This is not standalone - there are other PR's tackling this feature. Tests are supposed to pass only after all subPR's are merged.

### DIFF
--- a/crates/utils/src/settings.rs
+++ b/crates/utils/src/settings.rs
@@ -1,0 +1,251 @@
+use crate::communication::consumer::{CommonConsumer, CommonConsumerConfig};
+use crate::communication::parallel_consumer::{
+    ParallelCommonConsumer, ParallelCommonConsumerConfig,
+};
+use crate::communication::publisher::CommonPublisher;
+use crate::notification::full_notification_sender::FullNotificationSenderBase;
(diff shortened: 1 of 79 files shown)


crates/utils/src/settings.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct MonitoringSettings {
+    #[serde(default)]
+    pub metrics_port: u16,
+    #[serde(default)]
+    pub stats_port: u16,
+    #[serde(default = "default_otel_service_name")]
+    pub otel_service_name: String,
```
frondeus, line 85: Is this used? I don't see any code in `utils::tracing::init()` to load this service name.

luke-biel: Good question. It was present in horust config, but I guess I've never asked myself a question where am I plugging it in.

esavier (approved):

PR #459, opened 2021-05-07
feat: Update helm depolyment (config-files part 3)

luke-biel:
This updates deployment (helm only) to use new env variables.

I would like to ask one question with this PR - does anyone use docker-compose any more? For anything except infra. If no, we can remove it from cdl.


smores56:
I think docker-compose is simpler than helm, and I have used them for running the Python tests, but not for anything else. I'm fine with removing them.

smores56 (approved):

PR #461, opened 2021-05-07
feat: Provide bare-bones toml configuration files (config-files part 5)

luke-biel:
This docs aren't finished, but are a place to start when configuring CDL


esavier (commented):
I would propose that to make the config universal across deployments, lets get rid of the duplicated ports, i.e. each service will have its own metric port and so on.

esavier (approved):

Issue #452, opened 2021-05-05
docs: discussion about deployment.


esavier:
We have to create an RFC proposing how are we gonna store deployment files/templates for CDL. We need to take epicli into account, so a tool that generates k8s config or pure k8s config is a requirement. We will most likely use this config for tests.

PR #485, opened 2021-05-17
docs: deployment rfc

luke-biel:
https://github.com/epiphany-platform/CommonDataLayer/issues/452 - foundation for discussion


docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,90 @@
...
+* optional
+* linux-only (uses linux api's incompatible with OSX and Windows)
+* bare-metal
+
+### custom bare metal
+* optional
+* configuration can be propagated to other forms of deployment
+
+## Proposed
+We revisit helm and publish a template where you plug in individual toml configurations.
```
frondeus, line 48: We should probably mention downsides of using HELM and especially k8s:

 - it slows down the development because of slow `docker build` times. (While it uses some cache, its still far from perfect bare-metal solution)
 - It requires a developer, tester to run k8s environment for example `minikube` - which might work - or not (especially on windows).
 
 However if HELM is going to use `toml configurations` and not ENV variables embedded into helm chart templates, then I'm fine with it. I can develop and test with the Horust and then simply do one check, using minikube to see if there are any deployment errors.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,90 @@
...
+With this combination we can manage local development and k8s deployments, and have just one place where we keep configs up to date.
+Docker-compose and horust can stay in separate repo, for personal use. We shouldn't require a PR to update those as well 
+(not to mention config files can mitigate the need to update).
+We shouldn't provide `infra` helm files - just CDL deployment.
+Infrastructure `docker-compose` could stay in separate repo for development use.
+
+# Database setup
+
+## Currently
+We have docker-compose postgres setup and schema-registry setup in two separate directories.
```
frondeus, line 61: Not only we have docker-compose postgres setup, but also there is `schema-registry/migrations` folder which is used by tests. Therefore you need to maintain SQL in two different places. Three if you count Horust separate repo, but Horust could use docker-compose files from main repo.

There is also another problem with SQL migrations nowadays. In our migrations we use `SET SCHEMA 'cdl';` and unfortunately, if someone is going to change `sqlx::query!`, then they have to face sad truth - `cargo sqlx prepare` do not know about schema, and it searches for tables in `public` schema. 

It means - every time I change something in `sqlx::query` i have to drop Postgres, comment `SET SCHEMA` line, start Postgres again, wait few seconds, run `cargo sqlx prepare`, fix all bugs, uncomment `SET SCHEMA`, drop Postgres, start Postgres again, run our application to test if it works.

Its ... cumbersome.

smores56: It might be in our best interest to do some templating to handle that. Not sure what the best solution is (for templating, there are plenty of options), but that's the "easiest" way to avoid this problem without using a whole new system. That, or we move the migrations into the source code, which can be done with SQLx.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,90 @@
...
+* bare-metal
+
+### custom bare metal
+* optional
+* configuration can be propagated to other forms of deployment
+
+## Proposed
+We revisit helm and publish a template where you plug in individual toml configurations.
+This should reduce helm complexity and ease maintenance.
+We will keep cdl-config up to date. A good idea may be using `schemars` crate, or writing our own version with TOML support.
```
smores56, line 50: The `twelf` crate could also work here: support for JSON, TOML, ENV, Clap, and even Dhall (which is how I found it). https://github.com/bnjjj/twelf

esavier: its kinda immature yet, but i like the concept, it is worth checking out

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: right now migrations are out of scope, do not even think about those,
we do not have people nor time to use those,
tests will be using their own ways to fill table etc. the rest for now should go to the docs (i.e. whats needed)

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Currently
+We have docker-compose postgres setup and schema-registry setup in two separate directories.
+A client has to download these files and modify them before being able to apply.
+
+## Options
+### Embedding db setup into CDL
+Tricky and hard to manage. We have to remember that CDL is not Postgres only. While this may be an option for `edge` and `schema` registries, this is out of question for repositories now.
+### Setting up directory with migrations
+We will need to have migrations per `app`, assuming each application can use separate database.
```
esavier, line 68: yes, but again, not now, not today, not this quarter probably

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,90 @@
...
+We need to pick a tool to run these migrations.
```
esavier, line 69: not now,

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,90 @@
...
+#### Druid
+Probably need to write this tool ourselves. Druid requires conversion from CDL schema to its internal format, and maybe,
+assuming `SR-less deployments`, this tool has to be usable without schema whatsoever.
```
esavier, line 72: druid is no longer on required list, we can as well archive it, tasks related to druid currently do not have any priority

esavier:
Currently, there are other problems related to deployment, 

we are changing CI to different one, that means that we have to do background cleanup for all deployment related tasks and tests, and slowly migrate from github actions to CBS in terms of testing (validation can stay here, i guess)

The other problem is how to provide deployments and how to integrate with CBS properly. I believe helm files will be in CBS repo, while developer deployments can be moved to separate cdl-deployment. 

Having config files now (which should be integral to this repo and kept here) and later proper config service, we can assume that deployments can be abstract over the cdl itself, i.e. its the config section that will matter to the deployment not the deployment files itself (helm/horust etc) 
i started working on splitting the docker-compose infra to separate components so that we can choose which components to start each time, and no infrastructure will be running unnecessarily. 
- for docker-compose both infrastructure and cdl config will be handled by companion scripts
- for horust, there will be available separate scripts that will run only desired infrastructure set
- for helm right now there is no solution right now, and as you mentioned this may be unnecessary, i believe this will have to be discussed, but i believe it will be left for user to set up proper infrastructure for deployment to run on, the choice will be to use either script (i.e. from horust deployment) or set it up themselves. Config files have to be edited on the user side, though.

As mentioned before, and as with infrastructure, proper table setup will be left for the user right now until we will be able to find a solution to handle that on our side.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: We talking "db-init-scripts" using word "migrations". We don't migrate user data yet with them. We change schema from 0 to x. But we have to think it out as we were migrating our internal structure to new one. We already have 3 applications to provision db with. It's easier to talk in terms of `migration` frameworks instead of assuming an user will provision db for themselves (because this still requires that user to get sql scripts from somewhere) or whatever else will happen.

luke-biel: Templating seems like heavy cannon for a simple solution. But being honest, flyway doesn't seem like lightest tool too.
We have to avoid having two or more places with sql scripts, and we should avoid having to use `SET SEARCH_PATH TO cdl` in them.
Even simple bash script may work now. We can embed schema selection in it, leaving `.sql` untouched.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: `config-rs` is basically `twelf` from what I see.
`schemars` is there to generate JSON schema automatically - from source code - so we have one place to maintain. I don't see twelf supporting it.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: That's db initialization under fancy name only.

luke-biel:
> As mentioned before, and as with infrastructure, proper table setup will be left for the user right now until we will be able to find a solution to handle that on our side.

Yes, but we can't just assume that user will get this table setup from looking at our code. We have to provide them with sql for that. Or some description. But I assume - .sql is easier.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: we forgot `helm` has `pgInit` which isn't up to date with `ER`

Kononnable: >We have to avoid having two or more places with sql scripts, and we should avoid having to use SET SEARCH_PATH TO cdl in them.

If we use custom schema name there should be easy way to change it - for postgres, mssql using custom(set by us) name probably will be fine, for example mysql doesn't allow to create multiple schemas in one db(schema name = db name) - and we should have the ability to use same db server for multiple cdl instances.

Kononnable (approved):

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Proposed
+We revisit helm and publish a template where you plug in individual toml configurations.
+This should reduce helm complexity and ease maintenance.
+We will keep cdl-config up to date. A good idea may be using `schemars` crate, or writing our own version with TOML support.
+We will provide basic config for `kafka + all repositories` and `rabbit/grpc with postgres`.
+With this combination we can manage local development and k8s deployments, and have just one place where we keep configs up to date.
+Docker-compose and horust can stay in separate repo, for personal use. We shouldn't require a PR to update those as well 
+(not to mention config files can mitigate the need to update).
+We shouldn't provide `infra` helm files - just CDL deployment.
```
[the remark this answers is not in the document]
luke-biel: Helm infra stays as is. Helm, as mentioned, is a baseline, and infra is not harming anyone for being present there. It makes this "initial setup", our testing etc. much easier.

PR #488, opened 2021-05-17
feat: Schema registry validates view's relation ID

### LINKED ISSUE #487: feat: Schema registry validates view relation id before upserting
Right now we accept all UUID for relation_id even if there are no relations in edge registry.
Schema registry should call edge registry and ask if relation exists.

This is a continuation of #435

frondeus:
Fixes #487


crates/edge-registry/src/lib.rs
```diff
@@ -131,6 +131,21 @@ impl EdgeRegistryImpl {
             .map(|row| row.get::<_, Uuid>(0)))
     }
 
+    #[tracing::instrument(skip(self))]
+    async fn validate_relation_impl(&self, relation_id: &Uuid) -> anyhow::Result<bool> {
```
smores56, line 135: From a naming perspective, validate is a verb that implies that the method will do something, not return info. Since this returns whether or not the relation exists, you should either rename it `relation_exists` or return `anyhow::Result<()>`, generating an appropriate error if there is no relation.

smores56 (changes requested):

esavier (approved):

crates/edge-registry/src/lib.rs
esavier: yeah, i agree with Sam, although its just cosmetics

esavier:
i would like this feature to be configurable
there is a plan to add configuration to each module, similar to MATERIALIZER_CONFIG for the materializer, so while we are adding validation we can add it as well, 
i.e. it will provide a way to turn the validation on/off on per-module base, in this case it would be per view,
proposed struct 
```
Relation { 
  relation_id: uuid,      
  parent_schema_id: uuid,
  child_schema_id: uuid,
  config {
    validate_schema: bool,
  }
}
``
similarly with objects_id for each relations edge

frondeus:
How about moving this configuration to another PR?

esavier:
> 
> 
> How about moving this configuration to another PR?
okay, sure i agree

esavier (approved):

smores56 (approved):

PR #489, opened 2021-05-19
fix: Repair docker compose

luke-biel:


deployment/compose/setup.sh
```diff
@@ -0,0 +1,65 @@
...
+  docker-compose -f docker-compose.yml up -d \
+    druid_coordinator druid_broker druid_historical druid_middlemanager druid_router
+fi
+
+sleep 5s
+
+docker-compose \
+  -f docker-compose.yml \
+  up -d \
+  schema_registry # object builder needs it to be running
```
frondeus, line 41: Object builder uses BB8 pool now, so its no longer true. We don't have to run SR first.

frondeus:
Besides that, LGTM, although it does not unify configuration files with HELM, but I feel it might be a bigger challenge than I thought. We would have to somehow set different `[services]` for HELM (cause different DNS) than for D-C.

frondeus:
One thought - what about bare metal? By removing `.cdl` folder do we say that D-C is the preferred way of running CDL on local machines? Should I maintain my own config files set for Horust/ raw `cargo run`?

luke-biel:
> One thought - what about bare metal? By removing `.cdl` folder do we say that D-C is the preferred way of running CDL on local machines? Should I maintain my own config files set for Horust/ raw `cargo run`?

`.cdl` is not being dropped, in this PR there are only minor tweaks to it. D-C is there only for legacy reasons

deployment/compose/setup.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+COMMUNICATION_METHOD=${1:-"kafka"}
+REPOSITORY_KIND=${2:-"postgres"}
+
+QS_TYPE="document"
+
+if [ "$REPOSITORY_KIND" == "victoria_metrics" ] || [ "$REPOSITORY_KIND" == "druid" ]; then
```
esavier, line 9: druid is non standard, it can be left out

deployment/compose/setup/configuration/default.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[druid]
+url = "http://router:8888/druid/v2"
+topic = "cdl.timeseries.internal.druid"
+
+[notifications]
+destination = "cdl.reports"
+
+[monitoring]
+metrics_port = "0" # Random
```
esavier, line 36: i think we cant always leave status ports as random, i.e. i would prefer some specific port over random one, but it have to be different for each service
all deployments can use status endpoints for service monitoring, same with user. Probably also there will be tests for those later so letting those be random doesn't make sense without config service (i.e. place to notify about those ports)

deployment/compose/setup.sh
luke-biel: I wanted to cover it for as long as we don't delete it

deployment/compose/setup/configuration/default.toml
luke-biel: Yup, we can't always, but at this point, we weren't doing much with them on docker-compose anyway. If an user wants to turn on metrics/status they can change these in their local workspace

luke-biel: I've made metrics and status a static port in d-c by default

frondeus:
> > One thought - what about bare metal? By removing `.cdl` folder do we say that D-C is the preferred way of running CDL on local machines? Should I maintain my own config files set for Horust/ raw `cargo run`?
> 
> `.cdl` is not being dropped, in this PR there are only minor tweaks to it. D-C is there only for legacy reasons

Sorry, I don't know why I thought you removed whole folder

frondeus (approved):

PR #507, opened 2021-05-24
chore: Extract metrics & settings utils

frondeus:


crates/command-service/src/args.rs
```diff
@@ -3,7 +3,7 @@ use crate::output::OutputArgs;
 use clap::Clap;
 use thiserror::Error;
 use url::Url;
-use utils::metrics;
+use metrics_utils as metrics;
```
smores56, line 6: I'm not sure if we should do this. It's inconsistent with other usage. I see the benefits, but also it seems like it is done just to avoid changing the metrics definition in multiple places.

frondeus: We have to, or at least reexport `metrics` (the dependency, not our crate).
It is because `counter` macro expands to sth like `metrics::internal_function`, therefore callsite of the counter needs `metrics` in its scope. The other option is to add `metrics` dependency to every dependent crate but I wouldn't recommend it.

smores56 (approved):

PR #510, opened 2021-05-26
feat: PUE relations

luke-biel:
Functionality is here, but I'll need to work on optimizations in separate PR.

It would be good to extend each `Request` with information whether data came from `ER` or `CS`, because `ER` notifications guarantee that an uuid is tied in relation. But I'm leaving that out for now as well.

Optimizations that are missing out in this: https://github.com/epiphany-platform/CommonDataLayer/issues/513


crates/schema-registry/src/db.rs
```diff
@@ -185,6 +188,43 @@ impl SchemaRegistryDb {
...
+            filters as \"filters: _\",
+            relations as \"relations: _\"
+             FROM views"
+        )
+        .fetch_many(&mut conn);
+
+        Ok(stream
+            .filter_map(
+                |res: Result<Either<PgQueryResult, FullView>, sqlx::Error>| {
+                    let view = res.unwrap().right().unwrap();
```
frondeus, line 210: Is this unwrap safe? If so, could you write a comment explaining why?

crates/schema-registry/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![feature(drain_filter)]
+#![feature(async_stream)]
```
frondeus, line 2: Is this needed? Cant we use futures::Stream?

frondeus (approved):

PR #519, opened 2021-05-31
chore: Update dependencies and add miri testing

luke-biel:


luke-biel:
I changed semversions of crates to full ones via `cargo upgrade`. I assume automatic tool > our standards.

luke-biel:
As addition I'm dropping dependabot. I can set a reminder for every monday and do better at updating our deps than it does, from looks of this pr.

smores56:
So the changes themselves look good, and the removal of the dependabot is a welcome change, but I don't think we should include the patch version on all of our dependencies. It seems like it will just add a lot of annoying work when we can just leave the minor versions as-is and run `cargo update` to the same effect once a week.

luke-biel:
> So the changes themselves look good, and the removal of the dependabot is a welcome change, but I don't think we should include the patch version on all of our dependencies. It seems like it will just add a lot of annoying work when we can just leave the minor versions as-is and run `cargo update` to the same effect once a week.

Well, I agree partialy, as we initially wanted to use `major.minor` only, but this is automated via `cargo-upgrade` command (part of `cargo-edit` https://github.com/killercup/cargo-edit). I'm not sure we want to go into every file and change deps manually one by one - that's error prone. So I'd stick with whatever format `cargo-edit` choose.

.github/workflows/test-unit.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           profile: minimal
           override: true
-          components: clippy, rustfmt
+          components: clippy, rustfmt, miri
```
Kononnable, line 35: How is it related?

I think PR name needs updating to include all major changes.

Kononnable (approved):

.github/workflows/test-unit.yml
luke-biel: It's a minor change that didn't deserve separate PR, but here you go, changed the name

PR #520, opened 2021-06-01
chore: Remove lru_cache traces from data-router

luke-biel:
For some reason it wasn't used so I'm removing it now.

### DIFF
--- a/crates/data-router/src/handler.rs
+++ b/crates/data-router/src/handler.rs
@@ -1,12 +1,5 @@
-use std::sync::{Arc, Mutex};
-
 use anyhow::Context;
 use async_trait::async_trait;
-use lru_cache::LruCache;
-use serde_json::Value;
-use tracing::{error, trace};
-use uuid::Uuid;
-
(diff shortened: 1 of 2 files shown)


Kononnable:
@smores56 Can you take a look?
The only other place I see cache being used(after changes in #501) is `get_schema_insert_destination` which itself is not used anywhere. Did we just disabled cache 'by mistake' or are we using something else now?

smores56:
> @smores56 Can you take a look?
> The only other place I see cache being used(after changes in #501) is `get_schema_insert_destination` which itself is not used anywhere. Did we just disabled cache 'by mistake' or are we using something else now?

Yeah, this looks like the cache was disabled by mistake. In `crates/data-router/src/handler.rs`, the `route` function should be using the cache instead of loading from the SR every time. This PR should just use the cache and be renamed to "Properly User Cache in Data Router", or something to that effect.

https://github.com/epiphany-platform/CommonDataLayer/blob/bb2bc9baf6b092ffe307cdb13b75bfaf5fed84b3/crates/data-router/src/handler.rs#L129

PR #521, opened 2021-06-02
feat: cdl view_id and relation_id

luke-biel:
Allow to specify view_id and relation_id explicitly when adding new entities to db


Kononnable (approved):
Code looks fine, PR title is somewhat cryptic and probably not according to guidelines(no verb)

I'll ask to wait with merging to a meeting later today, so we make sure that this is the way we want CDL to go.

PR #546, opened 2021-06-08
docs: Feature List update

### LINKED ISSUE #545: docs: update feature list

esavier:
fixes #545


docs/features/index.md
```diff
@@ -10,30 +10,39 @@
...
-| CDLF-00010-00 | Protocol Versioning             | TechSpec | -----       | [RFC](https://github.com/epiphany-platform/CommonDataLayer/blob/develop/docs/rfc/CDLF-00010-00-rfc-01.md)                             |
-| CDLF-00011-00 | Basic Materialization           | TechSpec | -----       | [RFC](https://github.com/epiphany-platform/CommonDataLayer/blob/develop/docs/rfc/CDLF-00011-00-rfc-01.md) [DOC](./materialization.md) |
-| CDLF-00012-00 | Edge registry                   | RFC      | -----       | [RFC](https://github.com/epiphany-platform/CommonDataLayer/blob/develop/docs/rfc/CDLF-00012-00-rfc-01.md)                             |
-| CDLF-00013-00 | Materialized views              | RFC      | -----       | [RFC](https://github.com/epiphany-platform/CommonDataLayer/blob/develop/docs/rfc/CDLF-00013-00-rfc-01.md)                             |
-
+* in case there is a developmnent release we will mark it as XXX-rc, those are not fully fledged releases, a lot of the things can be broken there or undergoing testing
```
luke-biel, line 13: One thing, after we release 0.4-rc, i believe next release should be either 0.4-rc2 or 0.4. As we are pre-1.0 we could be fine with skipping rc part whatsoever and releasing just 0.x versions.

esavier: yeah thats the thing, probably there will be no 0.4.xx(-rc?) at all, more about this later

Kononnable (dismissed):
I have no idea if we have any criteria on how this should be approved or not(is this approval only needed for technical reasons), so approving. If this should be reviewed in some specific way someone who knows how should do it.

Few things about that change in general:
- changes view is a mess - every line was changed even if nothing valuable was added(probably spaces) - can't really be used in any useful way
- there is no `Abandoned` state mentioned anywhere other then in the table
- same goes for `Request`
- there is no 0.4.0 version yet, I think filtering is also not merged(nor fully tested) so I don't think it should have assigned version number just yet
- in general I'm not sure if we can call multiple things related to materialization a separate features - it might be just problem with naming though
- some features are not familiar to me even if I can say I know CDL rather well - e.g. Typeless Routing, Object-side configuration - without something which (even briefly) describes it I think it doesn't make much sense. I'm also not sure those 'planned' features were discussed with a team(they probably should be before they enter official list).

docs/features/index.md
esavier: oki, so we need to make 1.0.0 release near the end of the month, but 0.4.0-rc is gonna happen in this or next week.
Afterwards we will switch to -rcXX release strategy, i.e. 1.1.0-rc1, 1.1.0-rc2 etc,
major releases will be done quarterly, in between we have to make testing releases for people who want to try it

esavier: Does that solve it for you?

frondeus:
> every line was changed even if nothing valuable was added(probably spaces) - can't really be used in any useful way

You can tell GitHub to ignore whitespaces

docs/features/index.md
luke-biel: Yeah

frondeus (approved):

PR #548, opened 2021-06-10
docs: RFC: Typeless Routing

esavier:
CDLF-00015-00-rfc-01


docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,198 @@
...
+Currently, we can assume two scenarios:
+1. Absence of Schema Registry
+2. Limited presence of Schema Registry
+
+In the second case, assuming initial configuration or "mapping" of the repository_id to routing information can be provided from a different source than the Schema Registry, There is no issue in terms of routing.
+However, in the first case, without SR there is no possibility of providing advanced features like Materialization.
+
+## Proposed Solution
+
+### Configuration Service
```
frondeus, line 102: As an alternative solution I'd propose static mapping loaded from a configuration file (which in fact is already working). It means that schemas have to be known before startup of CDL but gives fully schemareg-less and cfgservice-less environment.

frondeus: Is this elegant solution? Probably not, however, we might need to ask whether that's the case (static configuration).

It resembles me a routing in proxies like Traefik or Envoy where you can either use dynamic routing (using k8s or docker dynamic service search capabilities) or static one, where proxy loads static yaml file.

esavier: Static routing may be problematic since:
- whole deployment will have to synchronize using the same file, that means there will be no update without restarting the whole deployment
- configuration cannot be changed without restarting the deployment

given that, i do not argue for or against it

Initially, the idea was to split this to two subtasks, one would introduce the feature without cfg-service, with some changes to schema-registry, while the second part would be done after that and splitting routing information from SR to CFGS.
This means changes to most of the services that is using SR to get routing info.

frondeus: @pljadru

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+##### Terminology
+* CDL - [Common Data Layer Project][cdl-project]
+* DR - Data Router, an CDL component responsible for ingesting and routing initial messages.
+* SR - Schema Registry, an CDL component responsible for keeping information about the type of the object conveyed inside the message.
+* User - user of the CDL. In this case, the user knows how the CDL works, and is assumed to have access to the API and code unless stated otherwise.
+* Message - (abstract) message sent to CDL for processing
+* Breaking Change - change in behavior, or API of the CDL that may result with system breakage on release update.
+* MD - man-day - amount of work completed by one developer in one work day.
+* Routing Information - information necessary for DR to successfully deliver the provided message to its destination. This includes, but is not limited to, addresses, ports, connections, metadata associated with ownership of the message, and abstract connection medium used by a specific repository handler.
+* CFGS - Configuration Service. Currently, with abstract behavior until a proper RFC will be concluded.
```
Kononnable, line 32: Please remove entries which are not used inside RFC text(e.g. MD).
It looks like a copy paste and waste of readers time if there are terms which are not even related to this RFC.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+
+##### Features:
```
Kononnable, line 34: Probably should be `related features` or something similar.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+        data: { any valid json },
+    }
+```
+
+v1.0 Input Format - Updated CDL message format to accommodate Typeless Routing.
+```
+    {
+        configuration: {
+            validation: bool,
+        } <optional>
```
Kononnable, line 57: How is validation related here?

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+###### Major concerns:
+There are a lot of changes to be made regarding this feature, and it has to be thoroughly tested before releasing. Total overhaul of how the messages look will force us to document all the changes and made them clear for the users. In addition, this Feature depends on the Configuration Service that currently lacks RFC and documentation (in R'n'D state), and Object-side Configuration (which is in Suggestion State)
+
+###### Notes:
+No testing nor PoC was created for this solution, please treat it as being theoretical.
+
+## Further Considerations
+
+#### Impact on other teams
+This feature MUST be released as a "breaking change", since it will impact basic message format.
```
Kononnable, line 162: Adding optional field is not a breaking change.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+
+##### Limitations
+Fully deployed CDL means a set of default features and configurations, including, but not limited to: ordering, materialization, access groups, at least 3 standard and default databases (Postgres, VictoriaMetrics, and object repository).
+With introduction of this feature, not all of those features and sets will be fully working or accessible. It is crucial to provide sufficient explanation and documentation, to clearly communicate the limitations that this feature entails.
```
Kononnable, line 103: I think this should be described in RFC - or at least mentioned which features will be 'disabled' for such configuration.

Kononnable (commented):
After reading RFC I have a feeling that it is not really about typeless routing, but about configuration service. And those are two separated changes. There is also validation mentioned here which again is another feature, not strictly related(related only on implementation level).

As for what is supposed to be main change here - typeless routing. As mentioned before - no one provided a user story how this feature helps a user yet. There may be a proper use case, but without knowing it we have no clue what exactly is the problem we're trying to solve and if  we can achieve that in a better way.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: k8s is an environment where restarts are supposed to be cheap and easy, as every service is small and redundant. I don't understand our fixation on dynamic configuration.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+  CDL feature ID  : CDLF-00015-00
+```
+##### Abstract
+```
+  The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
+  NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
+  "OPTIONAL" in this document are to be interpreted as described in
+  RFC 2119.
+```
+[RFC 2119 source][rfc2119]
```
luke-biel, line 19: Dead link

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+1. Absence of Schema Registry
+2. Limited presence of Schema Registry
+
+In the second case, assuming initial configuration or "mapping" of the repository_id to routing information can be provided from a different source than the Schema Registry, There is no issue in terms of routing.
+However, in the first case, without SR there is no possibility of providing advanced features like Materialization.
+
+## Proposed Solution
+
+### Configuration Service
+Configuration Service must be able to deliver routing information or in other words, "map" repository_id to routing information. There are two ways of handling routing information in terms of Data Router: caching it and expecting updates, or actively getting the data from CFGS :
```
luke-biel, line 115: As this is not an configuration service RFC i don't think all CS use cases were considered. It would mean, that before implementing CS we'd have to create documentation for CS and discuss it.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+With introduction of this feature, not all of those features and sets will be fully working or accessible. It is crucial to provide sufficient explanation and documentation, to clearly communicate the limitations that this feature entails.
+
+Currently, we can assume two scenarios:
+1. Absence of Schema Registry
+2. Limited presence of Schema Registry
+
+In the second case, assuming initial configuration or "mapping" of the repository_id to routing information can be provided from a different source than the Schema Registry, There is no issue in terms of routing.
+However, in the first case, without SR there is no possibility of providing advanced features like Materialization.
+
+## Proposed Solution
```
luke-biel, line 112: I can consider scenario where limited cdl would be desired - no materialization included, data is ingested with routing information, user keeps track of it's location in repositories, can query it via QS (not QR, or via QR with same routing information provided), notifications work. I believe something like this was proposed in first place by management - was it a mistake or have they other requirements than we are considering in regards to this?

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Introduction
+
+#### Background
+It was requested from the CDL stack to be able to route messages based on metadata in specific cases, ignoring or substituting information related to schemas.
+Currently, CDL is using Schema to decide which database to route the message to. All routing information is kept together with the type information, allowing multiple components and services to get this information if needed from one specific place (which, for the record, is the Schema Registry). After implementing this feature, CDL will be able to at least route the messages without using schema registry to get routing information from (also known internally as "schemaless cdl").
+The Scope of this document encapsulates changes to the current architecture and impact to already present features.
+
+#### Assumptions
+- We have to assume that different configuration of the CDL may exist and may be used for purposes CDL was not explicitly designed to used for.
```
luke-biel, line 95: Should we consider cdl configurations that aren't officially supported or requested by clients?
We should have approach - don't break things you don't have to; but I don't want to be considering what whole open  source community may do with cdl, especially when we aren't informed about such use cases.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+- Considering updated CIM format to be implemented, schema_id must be present, while repository name is optional.
+- It may be necessary to make this feature dependent on Configuration Service, and also split 'type' information from 'routing' information in schema_registry, preferably moving it to the configuration service.
```
luke-biel, line 97: That's a proposed solution, not assumption.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+- (PULL) In case of actively pulling data from CFGS, additional, unnecessary traffic and latency may be observed, additionally putting both services to greater stress (CPU and memory usage), however it will allow for greater consistency of message delivery to the proper repositories.
+- (PUSH) In case of caching, DR have to keep all routing information available for the current configuration, additionally CFGS must have an ability to update and notify routing information stored in DR. Comparing to pulling, there is slightly greater chance of data being inconsistently delivered and relatively bigger memory usage in DR (caching), however system itself will be much more responsive, have lower latency (since there is no additional network request per each incoming message) and CFGS does not have to scale with DR to keep up with the network load.
+
+### CIM Message changes
+CIM message have to be updates as described in the "Formats" section of this document. Additional field with repository_id should be present, while type
```
luke-biel, line 120: Even CS has to route messages via `schema_id` - as it's CDL message type identifier. If there's `repository_id` in messages, we don't need cs, only routing table.

Should we consider scenarios where `SID1`, `RepoID1` -> `database1`, `SID1`, `RepoID2` -> `database2`?
Why user cannot use two `SID`s in this case?

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+- - - valdation must not be performed
+```
+
+### Schema Registry
+Changes in schema registry would be limited to moving all the routing information from the Schema type, and moving those in some form of (list of mappings) to CFGS,
+no additional, high-level changes are necessary.
+
+
+### Services dependent on Routing Information
+Services like for example Query Router have to be updated as well, to accommodate new data source.
```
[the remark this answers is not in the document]
esavier: why i did not wrote it here is because i do not know the details and it was pretty hastily written. 
i assume that temporary SR will have to behave like CFGS but of course in limited capacity. I assume that simple api to ask for specific repository_id should be added (with cache) and all services that needs routing info will have to use that instead of getting SR schema structure.

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+    - Document storage.
+    - Time series storage.
+    - Binary storage.
+* Performance testing with all the before mentioned cases:
+
+## Deliberation
+Since Configuration Service is still in PoC state, splitting this in two separate tasks is crucial.
+
+#### Out of scope
+* Any other, not mentioned features, were not taken into consideration in this scope.
```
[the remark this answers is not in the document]
esavier: partially it is kinda legal jargon, partialy we have a shitload of features related to data router - ordering, batching, few missing rfcs like all of the formats (xml, avro, messagepack etc) compression, encryption. I skimmed through them and decided that there is no impact to those, so those should not be mentioned and thereby are out of scope. 
please, decide if i should remove it or fill it up?

docs/rfc/CDLF-00015-00-rfc-01.md
```diff
@@ -0,0 +1,211 @@
...
+    - config unavailable, schema available,
+    - config unavailable, schema unavailable,
+* Responsiveness test:
+    - config update latency
+* Happy test:
+    - Proper behavior with proper values.
+* Format testing for each, readily available repository type:
+    - Document storage.
+    - Time series storage.
+    - Binary storage.
```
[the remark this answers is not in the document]
esavier: i think there should be tests even if now those won't do much between formats, just for sake of being covered in case of future changes. I do not know how to handle binary storage since its literally missing (maybe we can drop it actually) byt about the others testing should stay here.

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: cool,
when? its now even for this quarter

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: because it was copy-paste
and its the only one line that is not relevant for the rfc

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: right now none, thats why i did not mentioned any

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: CS is not impacted by this change

esavier: also CS was implemented last year on the beginning of the project

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: its no longer a case, SR must be there, name mapping is just another "friendly name" that will be used instead schema, but schema can not be absent.
i materialization is not there its not there but tits working with this feature if needed

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: all possible usecases are valid, including putting it into istio or other service meshes, using vpns, scaling or different communication methods. Those were taken into consideration but so far does not impact the behavior.

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: thats the only solution, stop with static routing
levels of configuration must be separated based on access levels,
even partially implemented feature is better with separate than with static config

docs/rfc/CDLF-00015-00-rfc-01.md
esavier: CS is command service, it does not have to route anything,
again, no static configs, if you want to half-ass the solution then do not do it at all.
requirements changed and we are delivering whatever is to deliver in shortest possible time. currently only DR need to know routing info, QS,QR, and materialization modules are left out of scope for Q2, SR have to get new searchable field, which is "repository" name and be able to search either by that or by schema_id, in case of repository search, it can return any schema_id since only required value for now is routing info, (in that case schema_id won't be used) there is no impact, and implementation is fast and easy.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: The idea behind typeless routing is, I believe, getting rid of SR. Unless we hear different, I'm not sure if this will be accepted.

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: I meant, this doesn't belong in this section

docs/rfc/CDLF-00015-00-rfc-01.md
luke-biel: I meant `ConfS`.
I don't see benefit from having `repo_id`. `schema_id` was always identifier for `type + repo`, even if implicit. this means we already have such feature implemented. if we want to add searches via `schema name`, that can be done. it's same amount of work.

PR #555, opened 2021-06-15
chore: drop gh-pages in admin-panel

luke-biel:


luke-biel:
This ill fix security vulnerability permanently. Plus we aren't deploying to gh-pages.

esavier (approved):

PR #557, opened 2021-06-15
feat: plug web admin into sr

luke-biel:
I've limited admin to only view schemas for now. This fetches data from `api`, and assumes `sr` running. To show anything we have to fill in `sr` via other means.

I've also ran `prettier` on all svelte, ts and js files


web-admin/.gitignore
```diff
@@ -1,4 +1,3 @@
 /node_modules/
 /public/build/
-
-.DS_Store
+src/generated/graphql.ts
```
Kononnable, line 3: Should we ignore that file if it's generated only on running specific command(as I understand from readme), not executed by build command?

web-admin/src/sample-data.ts
```diff
@@ -1,146 +1,39 @@
...
-    objectId: "a96d4792-a68f-4262-b0b5-aa4341f80ca6",
-    data: { some: "data 6" }
-  },
-  {
-    schemaId: "29ff99d5-10ec-4b00-be71-a16380fad3a9",
-    objectId: "946afff9-126f-4a7c-bf26-7869021efda5",
-    data: { some: "data 7" }
-  },
-];
+import type { InputMessage } from "./generated/graphql";
```
Kononnable, line 1: Is this file still relevant?

web-admin/src/stores.ts
```diff
@@ -1,22 +1,19 @@
...
-export const darkMode = writable(
-  localStorage.getItem("dark-mode") === "true",
-);
-
-darkMode.subscribe(isDarkMode => {
-  localStorage.setItem("dark-mode", JSON.stringify(isDarkMode));
-});
+import { writable } from "svelte/store";
+
+const DEFAULT_GRAPHQL_ENDPOINT = "http://localhost:50106/graphql";
```
Kononnable, line 3: We probably need to take it from an env var or some similar mechanism, but this can be done in the future.

web-admin/.gitignore
Kononnable: Yep, currently build script for clean repository fails because lack of that file.

web-admin/src/stores.ts
luke-biel: Agree, let's leave it for later

web-admin/src/sample-data.ts
luke-biel: Removed it

web-admin/.gitignore
luke-biel: I can agree on keeping this file in repo. We will need to add action checking whether our codegen is up to date at some point. I don't think we even do that for rpc.

Kononnable (approved):

PR #558, opened 2021-06-16
test: object builder e2e tests for materialization phase I

Kononnable:
- e2e tests for object builder, on demand materializer - materialization phase I
- fixes to k8s(still some more fixes there needed)
-  enable jaeger tracing for k8s deployment
- use same pg migration files for k8s and docker compose


e2e/src/api.rs
```diff
@@ -0,0 +1,206 @@
+use std::collections::HashMap;
```
luke-biel, line 1: I'm worried about using api for tests - we are currently storing it, but it doesn't have to be up to date nor maintained actively. Which means these tests may randomly break.

Let's just make sure everyone in team is okay with moving api to our `"officially supported"` tier

luke-biel (commented):
It looks good, just make sure that you address my comment, if it's okay with team I can just approve.

luke-biel (approved):

frondeus:
Adding e2e crate has one major disadvantage. I can no longer do `cargo test --all --all-features` on my local machine because e2e runs first and then fail when I have no environment ready

frondeus:
> Adding e2e crate has one major disadvantage. I can no longer do `cargo test --all --all-features` on my local machine because e2e runs first and then fail when I have no environment ready

One solution would be to add some kind of env variable required to run tests, like `CDL_E2E` that would control it. It would be set for `e2e/test.sh` and on CI but would be missing by default for casual run.

Kononnable:
`cargo test --all --all-features --exclude=e2e` should work properly.
I don't know if we want to keep this e2e setup, or migrate it to something else in the future. For know I don't know how to easily(with low cost) to make it better - adding env var is an option, but every test would have to check it and return Ok() even if test was skipped(I'm not sure if we can skip test based on env var).

frondeus (approved):

PR #563, opened 2021-06-18
fix: Use view plan to filter objects from Query Service

frondeus:


crates/object-builder/src/view_plan.rs
```diff
@@ -54,6 +59,8 @@ pub struct ViewPlan {
     pub(crate) missing: HashMap<ObjectIdPair, Vec<usize>>, // (_, indices to unfinished_rows)
     #[serde(skip)] // Serialize is used only for tests, we dont need to assert view
     pub(crate) view: FullView,
+    #[serde(skip)] // Serialize is used only for tests, we dont need to assert view
```
frondeus, line 62:
```suggestion
    #[serde(skip)] // Serialize is used only for tests, we dont need to asset it
```

Kononnable (approved):

PR #577, opened 2021-06-23
feat: validation

luke-biel:
I know there's no rfc beforehand, but this introduces cdl-lite like validation, with additional separate service that can also do this validation. I guess it's good feature to add before Q2 ends, it wasn't much work, and we can rfc it later to discuss if we need all of it or only parts / change some behavior.


esavier:
Validation is a feature that needs either doc or rfc to not make mistakes like in case of ordering.

luke-biel:
> Validation is a feature that needs either doc or rfc to not make mistakes like in case of ordering.

It there: `docs/features/validation.md`. I guess I can expand on it a little, but a rfc is of no use here - we have no specs from any of the clients, I can only describe what we are doing, not what we are supposed to do.

esavier:
> > Validation is a feature that needs either doc or rfc to not make mistakes like in case of ordering.
> 
> It's there: `docs/features/validation.md`. I guess I can expand on it a little, but a rfc is of no use here - we have no specs from any of the clients, I can only describe what we are doing, not what we are supposed to do.

RFC is a process in which you describe problem, solution and implementation routes. Validation is a feature, not a simple expansion/bugfix. I am also not going to block this per se, as usually, you will do whatever you want either way, but be mindful that RFC process will be official by tomorrow ( as by PO, not by me ).
Have fun :)

PR #589, opened 2021-06-28
feat: hide e2e tests under feature flag

### LINKED ISSUE #584: bug: Cargo test require third party services to be run
cargo tests requires graphql for its testing which is not started during testing itself. Either make it so `cargo test` does testing properly or drop third party tooling from the testing

Kononnable:
Closes #584

### DIFF
--- a/e2e/src/lib.rs
+++ b/e2e/src/lib.rs
@@ -1,13 +1,13 @@
-#[cfg(test)]
+#[cfg(all(test, feature = "e2e"))]
 mod api;
-#[cfg(test)]
+#[cfg(all(test, feature = "e2e"))]
 mod object_builder;
 
-#[cfg(test)]
+#[cfg(all(test, feature = "e2e"))]
 const POSTGRES_QUERY_ADDR: &str = "http://cdl-postgres-query-service:6400";
-#[cfg(test)]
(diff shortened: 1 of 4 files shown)


e2e/src/lib.rs
```diff
@@ -1,13 +1,13 @@
...
+#[cfg(all(test, feature = "e2e"))]
 const POSTGRES_QUERY_ADDR: &str = "http://cdl-postgres-query-service:6400";
-#[cfg(test)]
+#[cfg(all(test, feature = "e2e"))]
 const POSTGRES_INSERT_DESTINATION: &str = "cdl.document.data";
-#[cfg(test)]
+#[cfg(all(test, feature = "e2e"))]
 const POSTGRES_MATERIALIZER_ADDR: &str = "http://cdl-postgres-materializer-general:6400";
-#[cfg(test)]
+#[cfg(all(test, feature = "e2e"))]
```
frondeus, line 12: I might be wrong but instead of marking all items in the module you could mark module itself.
In the beginning of the file:
```
#![cfg(all(test, feature = "e2e"))]
```

Note the extra exclamation mark.

Kononnable: Yeah, silly me, thanks

frondeus (approved):

Issue #595, opened 2021-06-29
web-admin doesn't send version when inserting objects


luke-biel:
As in title, there's mandatory VERSION field in CDLIM, that wasn't propagated to web-admin

PR #597, opened 2021-06-29
fix: Ensure web-admin sends version when inserting data to data-router

luke-biel:
Closes https://github.com/epiphany-platform/CommonDataLayer/issues/595


frondeus:
As you are working on it could you please fix API as well? 
In `crates/api/src/schema/mutation.rs` we require versioning but for some reason, we don't use it and then we put hardcoded const.

frondeus (approved):

PR #601, opened 2021-06-30
chore: put cache in utils

luke-biel:
This is refactor I did while implementing validation. Basically, we had boilerplate for pulling data into cache in 3 places. I've merged them into something a little more condensed.

I've also found out I missed 2 routes in QR when implementing static routing. I've fixed my mistake.

Original PR for reference: https://github.com/epiphany-platform/CommonDataLayer/pull/577


crates/data-router/src/handler.rs
```diff
@@ -23,16 +23,15 @@ static CDL_INPUT_PROTOCOL_VERSION_MAJOR: u64 = 1;
...
     pub producer: Arc<CommonPublisher>,
-    pub schema_registry_url: Arc<String>,
     pub task_queue: Arc<ParallelTaskQueue>,
     pub routing_table: Arc<HashMap<String, RepositoryStaticRouting>>,
 }
 
 #[async_trait]
 impl ParallelConsumerHandler for Handler {
-    #[tracing::instrument(skip(self, message))]
+    // #[tracing::instrument(skip(self, message))]
```
Kononnable, line 34: ?

luke-biel: A mistake, fixed

crates/data-router/src/handler.rs
```diff
@@ -204,17 +191,15 @@ async fn route_static(
     Ok(())
 }
 
-#[tracing::instrument(skip(publisher))]
+#[tracing::instrument(skip(cache, publisher))]
 async fn route(
-    cache: &Mutex<LruCache<Uuid, String>>,
+    cache: Arc<Mutex<DynamicCache<Uuid, String>>>,
```
Kononnable, line 196:
```suggestion
    cache: &Mutex<DynamicCache<Uuid, String>>,
```
plus replacing ` route(self.cache.clone(),` to `route(&self.cache,` in two places.
We don't have to clone `Arc`. Not much but this saves us from unnecessary synchronization between different threads(cloning Arc).

Kononnable: Tbh. I think we can safely remove `Arc` now - it was needed earlier, but I don't see any more of it being cloned.

crates/data-router/src/handler.rs
```diff
@@ -204,17 +191,15 @@ async fn route_static(
...
     event: &DataRouterInsertMessage<'_>,
     key: &str,
     publisher: &CommonPublisher,
-    schema_registry_url: &str,
 ) -> anyhow::Result<()> {
-    let insert_destination =
-        crate::schema::get_schema_insert_destination(cache, event.schema_id, schema_registry_url)
-            .await?;
+    let mut cache = cache.lock().await;
+    let insert_destination = cache.get(event.schema_id).await?;
```
Kononnable, line 202: I'm not sure this is a correct behavior - we're keeping a lock while requesting (over the network) destination address. This will block the whole data router until a response is received - messages sent to different schemas should not be blocked by a request for another schema(unless the same order group id is used but this is handled elsewhere).

crates/materializer-general/src/lib.rs
```diff
@@ -152,18 +121,16 @@ impl GeneralMaterializer for MaterializerImpl {
             .parse()
             .map_err(anyhow::Error::from)
             .map_err(error_handler)?;
-        let view_definition = self
-            .get_view_definition(view_id)
-            .await
-            .map_err(error_handler)?;
+        let mut view_cache = self.view_cache.lock().await;
+        let view_definition = view_cache.get(view_id).await.map_err(error_handler)?;
```
Kononnable, line 125:
```suggestion
        {
            let mut view_cache = self.view_cache.lock().await;
            let view_definition = view_cache.get(view_id).await.map_err(error_handler)?;
        }
```
We don't have to keep mutex locked the whole time + same issue as with the previous comment. However, that might not be a problem with materialization yet - OB is still limited to process one view at a time(for now).

crates/data-router/src/handler.rs
luke-biel: Good find, didn't think about this case.
Difficult change - for me `get` should receive `&mut`, as it's, well, mutating state. But I can't achieve this from outside unless I change cache to be behind a mutex, or `handle` would have signature mentioning `&mut self` (which is impossible, cause &mut is exclusive). Thus I have to put mutex within the cache itself, and call `get()` with normal reference. This also forces me to clone, unfortunately, as guards are short lived. I've experimented with returning a reference, but that's more work than I'm willing to put into it at this moment. So I'm cloning internally. In practice, this doesn't change performance, as we were already cloning fetched data.
I guess there could be a better solution somewhere, but probably with mentions of `unsafe` or `static`.

luke-biel:
I've changed `lru_cache` as it was unmaintained.
Benches
```
test tests::bench_lru_insert                 ... bench:      52,968 ns/iter (+/- 5,991)
test tests::bench_lru_cache_insert     ... bench:     115,097 ns/iter (+/- 7,874)
test tests::bench_lru_cache_get         ... bench:       1,693 ns/iter (+/- 295)
test tests::bench_lru_get                     ... bench:         836 ns/iter (+/- 69)
```

crates/query-router/src/schema.rs
```diff
@@ -0,0 +1,34 @@
...
+impl SchemaMetadataSupplier {
+    pub fn boxed(schema_registry_url: String) -> Box<Self> {
+        Box::new(Self {
+            schema_registry_url,
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl CacheSupplier<Uuid, (String, SchemaType)> for SchemaMetadataSupplier {
```
smores56, line 21: Though this is a simple use case, I would rather us make a struct with the fields `query_address` and `schema_type` to avoid having a bare string thrown around.

crates/data-router/src/schema.rs
```diff
@@ -1,45 +1,30 @@
...
-        schema_id
-    );
-    cache
-        .lock()
-        .unwrap_or_else(abort_on_poison)
-        .insert(schema_id, insert_destination.clone());
+#[async_trait::async_trait]
+impl CacheSupplier<Uuid, String> for SchemaMetadataSupplier {
+    async fn retrieve(&self, key: Uuid) -> anyhow::Result<String> {
+        let mut client = rpc::schema_registry::connect(self.schema_registry_url.to_owned()).await?;
```
smores56, line 20: Did you intend to remove the `trace!`'s, or did they get clobbered. I'm fine without them, just wanna make sure it was intentional.

smores56 (changes requested):

crates/utils/crates/cache/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+    async fn retrieve(&self, key: Key) -> anyhow::Result<Value> {
+        (*self)(key).await
+    }
+}
+
+pub struct DynamicCache<Key, Value>
+where
+    Key: Eq + Hash + ToOwned<Owned = Key>,
+{
+    cache_supplier: Box<dyn CacheSupplier<Key, Value> + Send + Sync>,
```
Kononnable, line 30: Maybe it's about what someone is used to, but I think it's easier to read code without unnecessary boxing.
My proposal - plus it can be made more readable by creating type alias eg. `type SchemaCache = DynamicCache<SchemaMetadataSupplier, Uuid, String>`. 
https://gist.github.com/Kononnable/b7892073ce1dbd23a3a8c7f9269edccc

Feel free to ignore it if it's a matter of being accustomed to something in your opinion.

crates/query-router/src/schema.rs
luke-biel: Yeah, I can agree on that. Changed.

crates/data-router/src/schema.rs
luke-biel: Well, tracing requires addition fmt::Display or Debug trait to signatures for Key, but I think it's not a problem. I've reinstated tracing in lib itself.

crates/utils/crates/cache/src/lib.rs
luke-biel: I've switched to static dispatch. I'm usually on conservative side when Implementing things, cause `<>` can grow fast. But it wasn't case this time, so I have no problem changing it, and it should be ever so slightly faster.

I'm not a fan of enums if there are many branches, and i'm almost certain that this enum would grow to house at least one more CacheSupplier.

Kononnable (approved):

PR #596, opened 2021-06-29
docs: fix feature list

frondeus:

PR #603, opened 2021-07-01
chore: feature list update

Kononnable:
continuation of #596

Introducing materializers isn't a feature on its own, it's part of CDLF-00013-00 Materialized views.
- described in the same RFC as materialized views
- materializers without materialization don't provide any value to a client(or CDL internals), the same goes for the other way around(materialization with no materializers - CDL will report errors). There is no known use case for materialization without any materializers nor for materializers without materialization.

### DIFF
--- a/docs/features/index.md
+++ b/docs/features/index.md
@@ -40,12 +40,10 @@
 | CDLF-00016-00 | Schema-Registry-less CDL deployment      | Ready      | 1.0.0       | [RFC](../rfc/CDLF-00016-00-rfc-01.md) |
 | CDLF-00017-00 | Materialization - Computation            | TechSpec   | -----       |                                       |
 | CDLF-00018-00 | Materialization - Materialized Types     | TechSpec   | -----       |                                       |
-| CDLF-00019-00 | Materializer - OnDemand                  | Ready      | 1.0.0       | N/A                                   |
-| CDLF-0001A-00 | Materializer - General                   | Ready      | 1.0.0       | N/A                                   |
-| CDLF-0001B-00 | Materialization - Notifications          | Ready      | 1.0.0       | N/A                                   |
-| CDLF-0001C-00 | Object-side configuration                | Suggestion | -----       |                                       |
-| CDLF-0001D-00 | CIM Object Valdiation                    | Suggestion | -----       |                                       |
-| CDLF-0001E-00 | Materialization - Relationships          | Idea       | -----       |                                       |
+| CDLF-00019-00 | Materialization - Notifications          | Ready      | 1.0.0       | N/A                                   |
+| CDLF-0001A-00 | Object-side configuration                | Suggestion | -----       |                                       |
+| CDLF-0001B-00 | CIM Object Valdiation                    | Suggestion | -----       |                                       |
+| CDLF-0001C-00 | Materialization - Relationships          | Idea       | -----       |                                       |
 
 ## References:
 [CDL RFC directory](https://github.com/epiphany-platform/CommonDataLayer/tree/develop/docs/rfc)


docs/features/index.md
```diff
@@ -40,12 +40,10 @@
...
-| CDLF-00019-00 | Materializer - OnDemand                  | Ready      | 1.0.0       | N/A                                   |
-| CDLF-0001A-00 | Materializer - General                   | Ready      | 1.0.0       | N/A                                   |
-| CDLF-0001B-00 | Materialization - Notifications          | Ready      | 1.0.0       | N/A                                   |
-| CDLF-0001C-00 | Object-side configuration                | Suggestion | -----       |                                       |
-| CDLF-0001D-00 | CIM Object Valdiation                    | Suggestion | -----       |                                       |
-| CDLF-0001E-00 | Materialization - Relationships          | Idea       | -----       |                                       |
+| CDLF-00019-00 | Materialization - Notifications          | Ready      | 1.0.0       | N/A                                   |
+| CDLF-0001A-00 | Object-side configuration                | Suggestion | -----       |                                       |
+| CDLF-0001B-00 | CIM Object Valdiation                    | Suggestion | -----       |                                       |
```
luke-biel, line 45: Is this about CIM format or CDL IM format? I don't wanna be nitpicky about changes that are partially unrelated to this PR, but I'm no longer sure.

luke-biel (approved):

docs/features/index.md
esavier: CDL input format or whatever mononclature is used rn
its a object validataion against schema

esavier (changes requested):
Both components you mentioned are features that are undocumented. Both components are features since they do introduce behavior that is noticeable by users. There is information what those are doing, but there is no documentation about those.

docs/features/index.md
luke-biel: In rest of the document it's referred as `CDL Input Message`, thus confusion.

esavier: when i was writing this it still was CIM, it can be changed but there is nothing official that was chosen

luke-biel: Hmm, can we agree on `CDL Input Message` maybe? @frondeus @smores56 @Kononnable . I can provide PR which changes it in all docs if we agree on that.
I understand in this case it's just legacy, but if I'm getting confused, chance is, others will too, so let's get over with it.

smores56: > Hmm, can we agree on `CDL Input Message` maybe? @frondeus @smores56 @Kononnable @esavier  . I can provide PR which changes it in all docs if we agree on that.
> I understand in this case it's just legacy, but if I'm getting confused, chance is, others will too, so let's get over with it.

The correct means is probably `CDL Input Message` defined in the glossary, abbreviated as `CIM` as we would abbreviate Data Router to DR.

luke-biel: I would skip abbreviations then, for now. Or change it to CDLIM.

smores56: Okay, no abbreviations, then.

frondeus:
> Both components you mentioned are features that are undocumented. Both components are features since they do introduce behavior that is noticeable by users. There is information on what those are doing, but there is no documentation about those.
> Additionally there is no developer documentation for object builder, and it was released.

Okay, this is the matter of "what is a feature". Is a new component a feature? Or the feature is "the new behavior of our system" and the component is only the implementation of this feature? Could this behavior be implemented in one single mono service for the sake of simplicity (but knowing its limitations regarding scalability)?

From my perspective, the feature is behavior, not a component. A component is only a tool used to deliver behavior. Therefore I agree with Paul.

However, I agree with you, Matthew, that both this behavior, its implementation, and its components are **undocumented**. Now what Paul proposed requires that all this documentation would be merged into one single RFC document. It's simply a matter of organizing stuff and being formal. No matter what option we choose, the same amount of work has to be put into this RFC[s].

Kononnable:
Put on hold until we define what "a feature" actually means.

luke-biel:
@Kononnable I don't think this is moving forward. Can we just close it?

frondeus:
> @Kononnable I don't think this is moving forward. Can we just close it?

Probably related task: #645

Kononnable:
Superseded by #645

PR #607, opened 2021-07-02
docs: codegen rfc

luke-biel:
I'm tackling here an issue of having to manually update two places when data schema changes.


docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+# Possible solutions
+
+## Generate rust code from specification files:
+
+There are options for generating rust code from proto (already in place) and OpenAPI.
+
+For generating structures needed for our internal and external communication, via Kafka and RabbitMQ, we could use [schemafy crate][schemafy-rs].
+It most likely needs to be done via `xtask`, similar to how we are generating things using [tonic][tonic-rs].
+
+As for OpenAPI, it would require us to switch away from using [warp][warp-rs] in favour of [hyper][hyper-rs]. Swagger can generate new crate in our repo,
```
esavier, line 48: what is swagger? a sentence or two of explanation?

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Generate rust code from specification files:
+
+There are options for generating rust code from proto (already in place) and OpenAPI.
+
+For generating structures needed for our internal and external communication, via Kafka and RabbitMQ, we could use [schemafy crate][schemafy-rs].
+It most likely needs to be done via `xtask`, similar to how we are generating things using [tonic][tonic-rs].
+
+As for OpenAPI, it would require us to switch away from using [warp][warp-rs] in favour of [hyper][hyper-rs]. Swagger can generate new crate in our repo, 
+that `QR` would use as a base.
```
esavier, line 49: ehm, newline by mistake? feel free to close this if its my mistake

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+
+For graphQL api - we would need to provide a `xtask` which pulls spec from running api instance. There's no easy way to generate rust code from existing api spec. I we were to stick with this approach - we'd have to write something in-house.
```
esavier, line 51: " if we were to... "

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+
+TOML configuration is a bit problematic. There are no known crates that can easily generate rust structs out from Toml config, and our structs
```
esavier, line 53: toml files are temporary until cfgs is introduced

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+We probably would need to produce an in-house solution for that problem.
+
+In general - it requires some significant work from team to switch fully to configs generating rust code and it can't be done easily in all of the places.
+
+## Generate configs from rust
+
+### Compiler pass way:
+
+We can create a binary that walks through our codebase and based on some criteria generates specs based on rust AST. It's doable, although not an easy task.
+More can be read searching for feature `rustc_private`, it allows users to use rustc as a library.
```
esavier, line 64: do i understand it correctly that we would hookup to the rustc and instrument it to generate some sort of ASN1 chunk as stage result?

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+`Configuration` is documented manually via [toml][configuration-docs] files compliant with [rust config crate][config-rs].  
+`gRPC` configuration is hand-written and [tonic][tonic-rs] generates rust files with code needed for client and server implementation; proto files are located in [rpc crate][cdl-rpc-rs].  
+`CDL Input Message` format is documented by hand in [docs][cdl-input-message]. At the moment of writing this rfc it's outdated.  
+`Notification Message` format is not documented to my knowing.  
+`graphQL` provides documentation via its endpoint and it's built into library we use. It may be wort considering to provide static graphQL spec in repository.
+
+# Problem to solve
+
+Maintaining documentation for formats used in communication within or with CDL is cumbersome.
+We need to research options of automating the process and validating its results.
```
esavier, line 37: i have a general question since i am unsure after reading the spec,
in which direction this document proposes an action to be taken?
(lets assume that .proto and other such specification files are counted as documentation) 
What i mean is: do we want to create code from doc or doc from code?
currently, for what i know, protocol files are a source not a destination of generators, but documentation about http/rest endpoints are generated internally inside source and no documentation is present besides manual one. Given so i am rather leaning towards generating doc from source if possible, since i am not aware of common spec format that different protocol code can be generated from (i.e. sounds like shifting the problem from one place to another)

docs/rfc/NAF-CDL_documentation_generation.md
frondeus: The question is - do we really need a file with graphQL API specification in our repo? I agree that this might be helpful - a user doesn't have to run CDL to figure out what API looks.

Maybe we could use either `xtask` or GitHub action (bot?) to run API on the CI, retrieve this dynamically created documentation and then prepare PR automatically?

We could use this approach - https://async-graphql.github.io/async-graphql/en/sdl_export.html

Also yes, I agree that there is no easy way to generate rust code from existing spec and I would not change it. It is too much work for too little gain.

Also that's why I advocate generating specs from code not other way around.

frondeus: Another question would be - do we need this kind of consistency if we choose to generate QR from spec. Cant rest stay as it is?

frondeus:
I'm not sure where to write this comment so it is not linked to any quote or line: I like what graphQL does - the fact that specification is the part of the API. That if you have somewhere for example running instance of CDL. Because maybe you are working in a team that has to use it and somebody created a developer environment with running CDL... then you don't need to check what version is running, search for specs in our repository. You can simply create a request to retrieve documentation from the living instance, and you are 99% that it is a good one (1% for errors and bugs :) ), and you can start experimenting with it.

I'd love to see something like that in our QR, therefore I think the argument that schema generation is mixed with "production code" can be thrown away. Because then, schema generation becomes "production code"

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -0,0 +1,111 @@
...
+I don't think it's wise to decorate function expressions with macros, and methods themselves don't provide all info out of the box.
+We would have to mark arguments of these methods with where does it come from. Some are provided by header, some are caches, some are bodies or query params. 
+If we wouldn't compromise speed and latency, I'd propose to review other options for servers in rust,
+as I believe they may be easier to integrate with `xtask` approach. Alas, there's a lot of uncertainty, and some of the research definitely requires a PoC.
+
+For configuration, it should be really easy to write these traits from scratch. It's ever possible to use `serde` own traits within our own (for defaults and whatnot).
+For JSONs, in form of `CDL Input Message` and `Notification Message`, goes the same. Just, we can use existing solution.
+
+`gRPC` may prove a challenge, more research is to be done, probably PoC. Tonic supports generation one way, 
+and I'm not sure how much of an effort would it be to replace generated files with something we'd be able to write manually and generate protos from it.
```
frondeus, line 83: There is also https://github.com/adamrk/tonic-rpc however it is not widely used

docs/rfc/NAF-CDL_documentation_generation.md
Kononnable: >Maybe we could use either xtask or GitHub action (bot?) to run API on the CI, retrieve this dynamically created documentation and then prepare PR automatically?

And merge PR just to make a new one pop up? I think PR should be complete when considered to be merged.
I see a few options how to handle it:
- keeping it in the repo(generating by the user)
- GH action generating the update and adding it as another comment to existing(open) PR
- not keeping it with codebase, storing it as release artifacts(I don't think we need them on daily builds)

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: If I don't put newline unfortunately markdown will consider this as one sentence. I wanted to stress that these topics tackle different cdl modules.

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: We don't need consistency, and I guess this will be one of topics today.

As for graphql spec in repo - I can see benefits - some IDEs provide validation based on that, we can read spec straight from file instead of starting an application to get it, I believe some errors also can be easier to find this way.

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: Even if, we will need a way to document config service, and if we reuse current `config tomls` code in there, even if tomls are unused in favor of some key-value store, we still should be able to generate specs from rust `struct` derives.

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: More or less yes. You get rust AST with expanded macros and imports, similar to ` rustc -Z ast-json filename`, just not in json. It's not the easiest solution to implement, but most versatile.

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: I also lean towards generating documentation from source, although I'm exploring all options here as to review if it's viable in all cases. Proto is toughest nut to crack.
As @frondeus mentioned, and as I wrote in one of the paragraphs, we may want to consider mixed solution. Main goal is just that we do as little manual work as possible.

> but documentation about http/rest endpoints are generated internally inside source

It's not, this is the openAPI I'm talking about here. You can view it in `crates/query-router/api.yml`, it's kind of json schema but for rest. We are currently updating it by hand, although there are ways to generate code from this spec. Unfortunately there are no known ways to do the opposite, generate API spec from rust code.

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: Hmm, I like it, though there's no `prost` support. And CDL is stuck with gRPC :/.

I was exploring options of prost support in tonic, and I think if we manage to implement `Deserialize` and `Serialize` on `prost::Message` it may be doable.

luke-biel:
> I'm not sure where to write this comment so it is not linked to any quote or line: I like what graphQL does - the fact that specification is the part of the API. That if you have somewhere for example running instance of CDL. Because maybe you are working in a team that has to use it and somebody created a developer environment with running CDL... then you don't need to check what version is running, search for specs in our repository. You can simply create a request to retrieve documentation from the living instance, and you are 99% that it is a good one (1% for errors and bugs :) ), and you can start experimenting with it.
>

As cdl developer, even if I cache these docs, they may get obseleted on different branches, and it's just a fuss to start cdl services, because you usually need kafka/postgres/zookeper aswel. That's where my preference of having them in repo comes from. If we xtask that, I'm pretty much okay with whatever we do.

> I'd love to see something like that in our QR, therefore I think the argument that schema generation is mixed with "production code" can be thrown away. Because then, schema generation becomes "production code"

I see one issue there, and it's optimization, of space and complexity. I would prefer to keep prod binaries as small as possible.

Although, I have to admit, I'm not sure if it's a real problem or just me being a purist.

docs/rfc/NAF-CDL_documentation_generation.md
luke-biel: I've added a note in glossary.

docs/rfc/NAF-CDL_documentation_generation.md
frondeus: Oh, I haven't noticed it, that's correct, prost is not supported. But maybe we could.. fork it if we are considering writing a custom solution.

frondeus:
> As cdl developer, even if I cache these docs, they may get obseleted on different branches, and it's just a fuss to start cdl services, because you usually need kafka/postgres/zookeper aswel. That's where my preference of having them in repo comes from. If we xtask that, I'm pretty much okay with whatever we do.

Maybe I wasn't clear enough - Id like to have both static spec stored in repository and also as response in living CDL instance.

docs/rfc/NAF-CDL_documentation_generation.md
esavier: thank you :)

docs/rfc/NAF-CDL_documentation_generation.md
esavier: what i mean that is this piece of document is relaying on something that will change as planned. 
would you mind reformatting it so that whats stated is about deriving and not configuration files itself? (just to be more precise, i do not have a problem with the idea itself) ?

docs/rfc/NAF-CDL_documentation_generation.md
esavier: okay, thats it, thank you for explanation.

docs/rfc/NAF-CDL_documentation_generation.md
esavier: okay, thank you for explanation.
You know my stance on the subject, given that i do not see any problem related to the generation in general.
From my perspective, this RFC can be closed when other issues with reviewers are resolved.

docs/rfc/NAF-CDL_documentation_generation.md
```diff
@@ -97,7 +100,14 @@ Some initial discussion happened on this topic, and it seems that team, in major
 
 # Conclusion
 
-TBD
+We've discussed this topic on 07.07.2021.
+Conclusion:
+* gRPC isn't our priority, and we are happy with current state; with `ap_pe_ce` being worked on, it's possible to revive this topic once it gets plugged into cdl.
```
Kononnable, line 105:
```suggestion
* gRPC isn't our priority, and we are happy with current state; with `ar_pe_ce` being worked on, it's possible to revive this topic once it gets plugged into cdl.
```

Kononnable (approved):

frondeus (approved):

PR #608, opened 2021-07-02
docs: RFC - Documentation restructure and RFC process

frondeus:
[Rendered](https://github.com/epiphany-platform/CommonDataLayer/blob/06ff7e8ae364f62cac2b055fac6da3cd98e24e58/docs/rfc/NAF-Documentation_and_RFC_process.md)


docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+If the deadline is met without conclusion, the author of the RFC should create a meeting and invite all interested in these RFC parties. Then during this meeting resolution has to be made.
+
+There may be exceptions to this rule - for example, when during a deadline meeting, everyone agrees to wait for new information from superiors, etc.
+
+### The process
+
+#### Draft
+The author of the RFC should create a new document in the `rfc` directory. This directory has a flat structure, and each RFC should have a filename in human-readable format with version prefix number: for example: `schema_registry_less_CDL_deployment_01.md`.
+
+Version prefix number is increased when RFC describes changes for existing RFC; for example, when there is `materialization_01.md` defining the first phase of materialization, and the author wants to describe the second phase, they should write `materialization_02.md`
```
Kononnable, line 156: I don't think materialization phases are a good example here.

I see three use cases here:
- (1) multiple RFC to a single feature
- (2) modifying the design of an existing feature
- (3) extending existing feature/adding new 'sub' feature

(1) - example: [CDLF-00011-00](https://github.com/epiphany-platform/CommonDataLayer/blob/rfc/Documentation_and_RFC_process/docs/rfc/CDLF-00011-00-rfc-01.md), [CDLF-00013-00](https://github.com/epiphany-platform/CommonDataLayer/blob/rfc/Documentation_and_RFC_process/docs/rfc/CDLF-00013-00-rfc-01.md)
Using similar names with different numbers sounds fine to me.

(2),(3) are similar, extending a feature is always in some ways modifying it. The distinction is hard to define sometimes.
For me, it's fine to use the same for (2) however, if RFC is mostly about including a new 'sub' feature I believe it might be good to use different naming. We may use parent feature as scope e.g.: `materialization_filtering`, `materialization_field_types`, `materialization_es_materializer`.
Why the difference? When we introduce a feature we don't know how big it will become over time. There may be a case when we introduce a feature and a year later half of CDL code is about that feature. We can split it(logically) into smaller scopes(logical, not related to implementation).

As for materialization phase II - tbh. it's really a term for tracking the implementation of 4 different features, which can be implemented on their own(independently of each other):
- filtering
- relation support
- computed values
- field types
It was designed in one 'document' due to lack of time, but they may as well be introduced as separate (sub)features.  Furthermore, because those features were introduced as a single term in 1.0.0 we didn't release materialization phase II even while the computed values feature is completed and there was no technical reason to not release it.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+Title           : Schema-Registry-less CDL deployment
+Category        : Feature
+Status          : Done(#123)
+Author(s)       : Wojciech Polak
+Team            : CommonDataLayer
+Reviewer        : CommonDataLayer
+Created         : 2021-06-24
+Last updated    : 2021-06-24
+Deadline        : 2021-06-26
+Version         : 1.0.0
```
Kononnable, line 172: When this field is supposed to change? As I understand RFCs won't be changed when they're in Done/Abandoned state - so only during implementation, or after it's agreed on, during implementation when something changes?

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+```` 
+# Front Matter
+
+```
+Title           : Schema-Registry-less CDL deployment
+Category        : Feature
+Status          : Done(#123)
+Author(s)       : Wojciech Polak
+Team            : CommonDataLayer
+Reviewer        : CommonDataLayer
```
Kononnable, line 168: Those fields will always have the same values and aren't really important.
It doesn't matter if RFC was created by A or B, only matters RFC text and who(how many) users need such feature.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+* Accepted - tracking issue should be created in CDL Team backlog. This tracking issue can contain more than one subtasks (sub-issues).
+* Done
+    * Update in `architecture` - all components that were changed should be updated, so this RFC won't be needed to understand how for example, Object Builder works. It should be done as a part of PR that has been merged.
+    * Update in `features` - if RFCs category is `Feature` - Update existing or create new feature.md and write a new description of what this feature does. In some cases, the author can selectively copy-paste it from RFC; however, the author must be careful. It should be done as a part of PR that has been merged.
+    * Update in `processes` - if RFCs category is `Process` - Update existing or create a new process.md and write a further description of how the process looks. In some cases, the author can selectively copy-paste it from RFC; however, the author must be careful. It should be done in PR that has been merged.
+* Abandoned - RFC should be updated with `Abandon reason`.
+
+It also means PRs with implementations should not be accepted and merged before merging documentation.
```
Kononnable, line 197: Before merging documentation or accepting RFC? Some changes may happen after the general idea is accepted during implementation(originating in implementation details). Not sure what tracking issue tracks in this case - implementation or just changing docs.

The question really is when should we do things like examples - it's part of documentation which is dependant on implementation.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+Request for Change might be:
+
+* Open (accepting new comments), 
+* Accepted (tracking issue has been created, RFC should also contain its ID: `Accepted(#123)`)
+* Done (implementation/process description has been merged to trunk branch, RFC should also contain PR ID: `Done(#456)`),
+* Abandoned (Team decided not to implement it)
+
+### RFC Deadline
+
+Each RFC must contain a deadline to speed up the process.
```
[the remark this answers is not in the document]
frondeus: We don't for now, I left it unspecified for the discussion

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+Version         : 1.0.0
+```
+
+# Glossary
+
+RFC - Request for Comments
+
+# Preface
+
+This RFC introduces a new process to keep our documentation clean and up-to-date. Both subjects - documentation restructure and RFC process is married and cannot be discussed separately.
```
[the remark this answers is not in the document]
frondeus: The best part is my grammar fixing program suggested `it` and I mindlessly accepted fix lol

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+Before the deadline is met, the author should remind all interested parties about it.
+
+If the deadline is met without conclusion, the author of the RFC should create a meeting and invite all interested in these RFC parties. Then during this meeting resolution has to be made.
+
+There may be exceptions to this rule - for example, when during a deadline meeting, everyone agrees to wait for new information from superiors, etc.
+
+### The process
+
+#### Draft
+The author of the RFC should create a new document in the `rfc` directory. This directory has a flat structure, and each RFC should have a filename in human-readable format with version prefix number: for example: `schema_registry_less_CDL_deployment_01.md`.
```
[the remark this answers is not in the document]
frondeus: Id keep it sorted in SUMMARY.md instead of numbers in filenames.

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: I agree, I missed that one

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: Tracking issues must track both implementation and documentation (if done separately).

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: Honestly, I dont care, we can remove them, we can keep them if someone wants them

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: Okay that example might be wrong but you got the idea so hurray :) ?

frondeus: I meant (2) and (1), i agree that (3) can be handled by another RFC. So more accurate example would be what you said about CDLF-00011-00, CDLF-00013-00 or for example versioning RFC where second RFC made first obsolete.

Kononnable (approved):

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+Status          : Done(#123)
+Author(s)       : Wojciech Polak
+Team            : CommonDataLayer
+Reviewer        : CommonDataLayer
+Created         : 2021-06-24
+Last updated    : 2021-06-24
+Deadline        : 2021-06-26
+Version         : 1.0.0
+CDL feature ID  : CDLF-00016-00
+Abandon reason  : Reason
```
esavier, line 174: i would also like discerning `abandoned` and `retired`
retired means that a feature ceased to exist or is no longer used in CDL
abandoned means there was a proposition, but it did not come into fruition

luke-biel:
> why rfc name have to be human readable?
@esavier
I didn't go through whole rfc yet, but I can back this particular thing up. It's convinience. I usually look for an rfc via filesystem rather than website, and it's difficult to distinguish which one is which. I also have no idea what all sections mean. Is it written anywhere? You've mentioned some format, `CDLF-XXXXX-rfc-YY.md`, so what does `CDLF-00016-00-rfc-01.md` first 00 mean?
Maybe we can find a way where we can filter through rfc's without the need to visit confluence (which is slow as fuck), and the problem solves itself?

esavier:
> > why rfc name have to be human readable?
> 
> @esavier
> I didn't go through whole rfc yet, but I can back this particular thing up. It's convinience. I usually look for an rfc via filesystem rather than website, and it's difficult to distinguish which one is which. I also have no idea what all sections mean. Is it written anywhere? You've mentioned some format, `CDLF-XXXXX-rfc-YY.md`, so what does `CDLF-00016-00-rfc-01.md` first 00 mean?
> Maybe we can find a way where we can filter through rfc's without the need to visit confluence (which is slow as fuck), and the problem solves itself?

@luke-biel 

CDLF-XXXX-YY-rfc-ZZ.md
XXXX - feature name
YY - subsequent rfc's for the same feature (supplementary ones)
ZZ - sequential version, for ones that obsolete one another (like with CDLIM version)

YY ended up not being used as much as i predicted, we can rid of it afaic
searching through file system is another problem, and i do not know how to solve it properly
of course user can have their own softlinks with different names in or outside the repo, also there are tools like grep that can find whatever document you need just by words used within. Usually, when i am working based on RFC i temporarily remember this number for the duration of the work and there is little to no reason to search for other ones (between 1 and 3 rfcs are being accessed at the same time at most either way) 
i am open for propositions but still the point about friendly names stands, a good example would be `schema-registry-less routing` which is neither describing the feature, nor is friendly to say out loud in conversations nor to reference in chats.
either way i am open for alternative solutions...

docs/rfc/NAF-Documentation_and_RFC_process.md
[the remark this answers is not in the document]
frondeus: In this response Im gonna use Rust RFC as an example: https://github.com/rust-lang/rfcs/tree/master/text

> why RFC name have to be human-readable? There is a title for it.
The title might not be enough. I agree that it is okay from reader's point of view. They have a statically generated website with summary.md and mentioned Feature list.

However, for me, a developer, it is often not enough. I want to quickly find an RFC on my computer using my editor/IDE. And while I could use the `Search` function by for example using `ripgrep` to find the title, or I could open Feature List and find the ID then type this ID to my `Find file` combo, but... it is simply cumbersome.

> currently rfc's bound to features are named with hexadecimal number, while feature list describes what is or what is not there. Feature list is an official list that is presented for 3rd parties on what features are actually in this version and in which state, and must be there.

Yes, that's true, because right now we have a tight join between features and RFCs. We did however at least once tried to use the RFC process to handle also processes and other, non-features like Commit Message formalization. Now I just try to solve that issue. I propose redefining RFC so it is more generic. However we could also agree to not use RFC for anything else, RFC would be tightened to features (technical, code).

Just to make it very clear: I don't want to remove the Feature list, nor I don't want to remove ID everywhere. I just want human-readable filenames.

> if a rfc is not technical rfc i.e. is not about the feature or code but about a process like this one, it was before named NAF-xxx (not a feature) not can be stored somewhere else so thats is not a problem.

That's a nasty exception for me. And that's what I described in the last paragraph. We treat features with higher priority than procedures therefore we tag NAF as something unusual. The procedure is here a second-class citizen. 

> CDLF-XXXXX-rfc-YY.md ment literally a serial number and ordered change i.e. 01, 02, 03 if any changes were made historically

Sure, I don't mind. We can even use plain integer like Rust does. Its simpler, you don't have to convert a number into hexes, it is even less prone for an error, and if we think that we will have more RFCs than Rust... that's a bit arrogant thinking. However, I didn't proposed changing the ID itself that much, because ... I don't mind. Could be done but I wont be mad if it stays as it is. 

Please keep in mind that I want to keep this order in some place, somewhat. We can do it by introducing `filename_YY.md`, we can reorder SUMMARY.md (however this now sounds like a bad idea, because it requires you to check SUMMARY.md).

> by naming it with human readable names you are introducing searching problems, since anyone can name it anyhing, sometimes even with something not representing the functionality itself (like SR-less routing, which is not about not relaying for routing on SR), I would like to have proper indexing and descriptions that are searchable and/or taggable

Okay, that's a big statement.

How can you prove that CDLF-XXXXX-rfc-YY.md format is better or searching when you explicitly said, that you need an index to find what feature has which ID?

I don't mind having this Feature list somewhere, for features, it is useful, especially that it stores not only links to the docs but also information in which version the feature was introduced.

But, I think raw indexes as a name are unreadable. I could however go for some kind of middle ground here. For example, in the Rust project, you have both an ID and human-readable name.

If you are afraid, that:
> since anyone can name it anyhing, sometimes even with something not representing the functionality itself

I agree that could happen, but. That's why we have the RFC process. That's why an author has to create PR and then wait for review. If you don't like how the filename is phrased, if you think that I would not be a good title, you can always comment on it, and require changes. The filename itself is also part of the process and the target for comments. 

Plus as I mentioned - Feature list is good, indexes are good.

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: > RFC can be made by 3rd party, and be non-(strictly)technical
reviewer can also be 3rd party sometimes,

Okay, makes sense it can stay. Is that however important? Honestly, I don't know. We know who wrote it (each commit has its author, git blame, etc.), we know who reviewed it (in PR we have git reviewer). Are we afraid that sooner or later we will have to move from Git or Github to some other source version control system and that's why we want to duplicate this information?

> whats the number in status? PR/issue link? RFC is detached from implementation, it only can be accepted or not. Implementation is a separate matter.

"Uh, when tracking issue has been created, RFC should also contain its ID." 
I know that you disagree with linking RFC to the implementation... But I kindly disagree with your disagreement. For the user it might be very handy to see, that for example there is a feature that he desires, it is described, he likes it and then he doesn't have to search for an implementation task, he can see exactly whats the status, whether we started implementing it or maybe it's already done.

Again, I took a cue from the RFC process in rust where there is a strong connection between RFC and its implementation. See -  for example https://github.com/rust-lang/rfcs/blob/master/text/2873-inline-asm.md

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: And why 6mths of dev time should stop us from linking issues with RFCs?

I agree that explaining reason could be done in more than one line.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+## What should we have
+
+RFC should only describe **change** to CDL. The change might be:
+* New feature
+* Modification of existing feature
+* New process
+* Modification of existing process
+
+It should not contain information about a single component (its current state, as this is the purpose of `architecture`).
+
```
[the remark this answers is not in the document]
frondeus: > sometimes there are rfc that are a result of a spike, rnd or whatever else, for example done in spare time or not bound to feature or "change". 
Okay, but still spike, R&D, or whatever else should have an issue created for that. However, I agree there might be an exception for Done status because not every spike or R&D can conclude with the implementation. 

On the other hand - maybe we should write R&D/spike results - comments, our insights into documentation, and therefore even Spike/R&D would result in Pull Request. That kind of Pull Request of course wouldn't change a single line of Rust code, but it would result in a new markdown document.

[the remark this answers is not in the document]
frondeus: > Additionally this is phrased like the creator can not mention anything about the component the change is related to.
Okay, I need to rephrase it. I don't mind if the creator mentions the current (in the time of writing RFC) state, I just want to distinguish between architecture and RFC, so we won't use RFC to describe how for example Schema Registry works in general.

docs/rfc/NAF-Documentation_and_RFC_process.md
esavier: then we are in disagreement,
first, rust-lang's rfcs or kafka KIP that you are bringing up are only one of many formats and do not hold much weight since its bound to personal preference and per-case usage. I am used to behavior displayed in, for example, ietf rfcs that are actually well written and structured. Link for examples:
https://www.ietf.org/rfc/
https://www.ietf.org/rfc/rfc5754.txt

Second, using friendly names is very short-sighted, in the future when we will have for example, lets say one hundred RFCS, you may or may not be able to remember what's the name was, and again, friendly name does not have to mean it is named properly. In that case, there are other options i mentioned - your own, local soft links, searching or grepping, or even making aliases in your favorite tools or IDE, like a plugin for fuzzy-finder. In that case, your  argument seems more like personal preference than a solution. The same argument had been presented when we were deciding about git-hooks, if you want those, do it in your workspace and do not force those on public.

docs/rfc/NAF-Documentation_and_RFC_process.md
esavier: I am not sure what problem is there to solve,
RFC may or may not be followed up by implementation, even better, some RFCs are a R&D with no implementation to be had. More important information is in which version RFC's result was introduced and thats what is happening currently in featurelist.md, if needed, such information can be also had there, however, i saw no industry standard RFC that kept that information inside the document itself, since both code and RFC are detached concepts.

docs/rfc/NAF-Documentation_and_RFC_process.md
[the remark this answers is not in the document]
esavier: research done against feature is usually technical one and result in document bound to technical concepts, like code, logic, algorithms etc. I may not be clear on that but what you are proposing is to smash every research into architecture documentation? If thats the case, just imagine what will the documentation look like when there are, lets say, 10 spikes on specific matter, and also how to archive it and present concise results for 3rd parties or new developers.

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: Okay now I can say that whatever you proposed is a matter of your preferences. And your preferences are not more important than mine. So sorry, but bringing "git-hook" as an argument does not work here.

Things like "these are actually well written and structure" is your opinion. So my example is not worse than yours.

But besides that.

> Second, using friendly names is very short-sighted, in the future when we will have for example, lets say one hundred RFCS.

Hundreds of RFCs are no problem here. Again you can see that Rust has now 497 RFCs and they are using the convention I proposed, and live and prosper well. 

Maybe if we go in thousands of RFCs sure, but then it's a bit arrogant. It feels like a premature optimization for a problem that might never come.

You don't like human-readable names? Sure, you can use your own options mentioned like soft links, searching or gripping, or even making aliases in your favorite tools or IDE, like a plugin that reads Feature list and points you to RFC that you need. In that case, your argument seems more like a personal preference, than a solution.

See? Exactly what you said. With human-readable names, you are not limited. You can always use this Feature list.
Plus we can meet in the middle ground here - let's make a rule that human-readable names must include the whole RFC title. We can accept spaces, or we could "URLify" them by replacing space with an underscore. Because then if you accept that the title is okay then you should also accept a filename as well.

docs/rfc/NAF-Documentation_and_RFC_process.md
frondeus: > RFC may or may not be followed up by implementation, even better, some RFCs are a R&D with no implementation to be had.
https://github.com/epiphany-platform/CommonDataLayer/pull/608#discussion_r665132793

frondeus: > I am not sure what problem is there to solve,

> For the user it might be very handy to see, that for example there is a feature that he desires, it is described, he likes it and then he doesn't have to search for an implementation task, he can see exactly whats the status, whether we started implementing it or maybe it's already done.

frondeus: Also define "industry".

For me industry standard is the standard of the language that we are using to write the code.
They attach issue link. So now what, are you going to tell me "my industry is better than yours"?

Cause in that way I feel this is irrelevant and a bit like a cargo cult.
Because sure, Rust does it but I'm not following it only because "that's the way", I have a real issue that I try to solve, and I feel that Feature list is here not a solution. 

I don't honestly understand why do you bother and why are you disagreeing. Only reason I see for now is "because IETF does not do it so I don't want it too".

frondeus: I don't understand how can you say that RFC and code are detached. RFC results in the code being written, or documentation being written, or insights from Spike/RFC collected. Otherwise, if RFC results in... nothing, then why bother?

Please convince me

docs/rfc/NAF-Documentation_and_RFC_process.md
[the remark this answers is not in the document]
frondeus: I think it is handy to share information about our thought and decision process, why we chose A over B.
We can create another category different than `architecture` for r&d and spikes. But yeah, making it public might explain to someone why the feature we build is in the state that it is. It can help that person to understand our design choices, to comprehend CDL better. And that is one of the purposes of having good documentation. 

An example that from my personal experience helped me understand the unknown for me architecture can be Epiphany itself: https://github.com/epiphany-platform/epiphany/blob/0992f4ed6588460981cae3ccb366655b0d89f376/docs/design-docs/cicd-server/comparision_cicd.md
Someone made an R&D, tested several different services, wrote a conclusion into the documentation, so then I could read it and understand better why they made that choice.

docs/rfc/NAF-Documentation_and_RFC_process.md
esavier: I do not see this as a healthy discussion, since you are following many logical fallacies,  argumentum ad nauseam to name at least one. For there is no middle ground, rather It's far and further ground from where i stand on this. To be clear, it sounds like you are stating that middle ground is anywhere in the domain of your proposition, and not actually a compromise.

So lets start with example of  RFC that did not nudge any code commit is a feature related to the message pack, and i believe further update will be required to understand the problem and find optimal solution for not only one protocol but also add ability to handle any or all protocols if needed. This is a example of R&D in which both logical and technological issues were pursued, checked, but there is no implementation and its totally find to do so in my opinion. Will the code be added in the future? maybe, maybe not. Maybe it will be abandoned in that stage.

Next one is definition of industry standard. True this is a very vague term, however it is commonly accepted that standard is something that multiple parties follow, not a singular project. Example of that is a MISRA C++ guidelines or JSF AV C++ standard to name at least two of those. Those are not arbitrarily created by one party to be used inside one project, but a joint venture between industries (as in industry standard) to follow some common path. I proposed IETF as a good example since they are much older and well-kept than any other RFC's available to general public. 
Also, i do not understand what you mean by "your industry is better than mine", as mentioned before, it is an example of well presented, common way of describing changes to general populace across different industries, both regarding internet communication, cryptography and general data representation and processing, to name a few.
Additionally, a deviation from norm is not a basis to make exception. In our case, deviation from norm that works in one project is not a basis to blindly implement it in this project.

I am puzzled about your last comment, since it seems that research on MessagePack that i've done on the topic was useless, despite RFC itself being the longest, where the problem was explored deepest so far.

Considering thread above about workarounds, if you wish to make my argumentation about those workarounds worthless then by conjunction you have to do it with your own too since you are making both of them to be based on personal preference. Difference is that linking with friendly-names does not work in reverse, i.e. there is no point in making unfriendly-named links, since there is already a mess in the directory. Additionally both you and me have to search what we want the same way in the end, i.e. not by name but by either keeping it open in background in the IDE or grepping the repo.
Moreover, its hard to miss that you didn't dispute the argument that names are not always representative of what is in the RFC itself.

Additionally, i do not have to convince you, you are not the sole deciding person in this discussion and have the same responsibility to convince everybody else to your position. I just stated that this solution is problematic and described what the problems are, hopefully conveying that both solutions are rather subjective and in the end we will decide by playing favors ¯\_(ツ)_/¯

frondeus: Thank you for your response.

I agree that my middle ground might be biased by my solution, and yes, that can be treated as a fallacy. I don't mind however if you present your middle ground so we can discuss it further and make something that would satisfy at least partially both sides.

About RFC that is not related to any code commit - message pack. I agree that this is a valuable research and the outcome may or may not be useful in the future. I could argue, however; whether the RFC status should be set to Done (or using another synonym - implemented). Perhaps we need another state, where we say "It is no longer open for comments, however, it is not implemented nor WIP", I guess the word `Accepted` could be used?
In that scenario there is no extra PR, there is no tracking issue, RFC stays in an Accepted state, waiting for its time, still being useful.

Industry-standard - okay, that made things much cleaner, now I understand your position. Thank you for explaining it to me. In that case, I don't mind if we agree to keep this format, however it is worth asking. Just asking - maybe we can do some differences to fit our problems better? For example, adding this implementation info.

As for being used by multiple parties not a singular project - here lives another fallacy, probably because I didn't understand. As far as I understand (please correct me if I'm wrong), you say that here in the CDL project we design another standard that can be implemented not only by CDL but also by other parties. And yes, that's true afaik and something I often forget. 

However I think (again, please correct if I'm wrong) but usually, such a standard is written by not only one side, but multiple parties, while here we as a CDL team decide and define the structure of the whole project. 

About deriving from the norm, I assume you see IETF as a norm here. Okay, point taken, but I disagree with "blindly implement in this project" - I do not blindly want to follow Rust's approach because it is Rust fresh and fancy. I hope I made it clear. I used it as a working example to back up my solution, which is based on other arguments mentioned before. I don't see anything wrong in including this example in the discussion and I believe we should not dismiss it prematurely only because it is not as old as IETF.

Workarounds - Yes, you made good points mentioning R&D, and yes I tried to fit them into this solution. That is the part of the process, to find out cases, and ambiguities and solve them, either by explaining how the solution can handle it or by changing the solution. It is indeed based on personal preference, and yours is different than mine. You don't have to change yours. If it is only a matter of personal preference perhaps we should choose one that would fit better the majority? 

Names not always being representative of what is in the RFC - in that case, it is the responsibility of the reviewer to notice it, and suggest a better title/filename. I am pretty sure during this process a good solution can be found.

Convincing - I'm sorry, that was poor wording. Perhaps a better one would be "Could you elaborate?". I agree that I am not a judge.

Thanks to your last comment your solution is (in my opinion) better described and rationalized. I think now we have better chances to exercise good judgment.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+features
+Purpose: a place for feature (behavior) description. Each feature should have at least one document describing it. Example: ordering.md, materialization.md, routing.md, CDLIM_versioning.md
+Change: Rewrite materialization.md by precisely describing what this feature is. Rest should be kept as it is, especially index.md with table linking feature with RFC and documentation.
+Used by: CDL user
+
+processes
+Purpose: a place for all our processes. For example: Commit message formalization or this RFC when it's going to be accepted.
+Change: New directory. There is no need for index.md because of SUMMARY.md
+Used by: CDL team
```
luke-biel, line 92: I think we should start indexing NAF, same as we do with Feats, for clarity.

Although, it's personal pref.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+
+deployment
+Purpose: Documentation of how to deploy CDL
+Change: No changes needed
+Used by: CDL user
+
+examples
+Purpose: a place for tutorials (for example, the one living right now in features/materialization.md, although it needs an update) and guides.
+Change: Move features/materialization.md, also maybe getting_started.md
+Used by: CDL user
```
luke-biel, line 82: This reminds me of Python Example, which is pretty much outdated. Should we delete it? Is this a good time to cover this?

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+
+## What we have now
+
+### RFC definition
+RFC is describing either the change or current state of the CDL. It might refer to other RFC and the current state of CDL. 
+There is no definition of what RFC is and what it should include.
+
+### Structure
+We keep all RFCs in the `rfc` directory; however, there is little to no resolution after we close it.
+Description of the components is mixed with RFCs describing features and processes, resulting in a strange mess where NAF-Commit_message...md lives next to `CDLF-000XX-rfc-XX.md`.
```
luke-biel, line 115: As for difficulty of finding an RFC via CDLF..., I agree with you that we should change it, and I like how rust team did it, just `$num-$name.md`. Few notes though:

We need to take our time picking names, so that there's no confusion.
We could go with old naming scheme if we introduced some bash search script.

docs/rfc/NAF-Documentation_and_RFC_process.md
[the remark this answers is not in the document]
luke-biel: I don't think there should be a deadline in doc. Recent events proven these are quite fluid.

I propose:
Author of an rfc keeps track of the rfc.
They look for good time to schedule discussion - if rfc/process is low prio, they WAIT
When there's time, meeting is scheduled, a comment in rfc is written about the meeting (optional)
This meeting is a deadline

Mainly, I'm affraid that low prio rfcs will push high prio ones if we will be too strict with this, and I don't see a benefit of deadline if there's no such strictness

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,223 @@
...
+CDL feature ID  : CDLF-00016-00
+Abandon reason  : Reason
+```
+```` 
+
+CDL feature ID is required only for the feature description. If RFC describes the process, this field should be omitted.
+
+Abandon reason is required only for Status: `Abandoned`.
+
+The author should commit the draft to a separate branch, preferably `rfc/<filename>`.
```
luke-biel, line 182: This is close to branch naming convention, we don't have one. I'd prefer to have `preferably ...` removed.

Kononnable:
Just to remind some things we decided on an earlier meeting(mid-march). Full notes on private keybase.

```
# RFC Meeting discussion

- 3 days is too short
- For 24 hours (give or take) before the meeting, no comments should be made
- People don't have to go to meetings if they don't have anything to add

# Approval of RFC's

- Can someone opt out of approving a PR?
- At least half of the team should approve of PR's (2 people for now)
```

Maybe it's worth discussing/adding it here/ making it official(the process of RFC, not just about documents).

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,218 @@
...
+### RFC Deadline
+
+Each RFC must contain a deadline to speed up the process.
+Before the deadline is met, the author should remind all interested parties about it.
+
+If the deadline is met without conclusion, the author of the RFC should create a meeting and invite all interested in these RFC parties. Then during this meeting resolution has to be made.
+
+The deadline must be set at least 3 days from the announcement. RFC author should schedule the meeting on the nearest possible date that is acceptable for all parties.
+
+Until the deadline, everyone is allowed to comment on the RFC. Participants should use the scheduled meeting only to decide when there is more than one strong opinion about the subject. All issues should be defined and discussed via GitHub comments. The deadline meeting should be relatively short if everyone agrees on the subject.
```
Kononnable, line 137: I would like to add here something that we introduced before - 'no comment phase'. To limit comments for a day before discussion - e.g. we cannot expect that comment written an hour before discussion will be read in time by everyone.

docs/rfc/NAF-Documentation_and_RFC_process.md
```diff
@@ -0,0 +1,218 @@
...
+```
+Title           : Schema-Registry-less CDL deployment
+Category        : Feature
+Author(s)       : Wojciech Polak
+Team            : CommonDataLayer
+Created         : 2021-06-24
+Deadline        : 2021-06-26
+Version         : 1.0.0
+CDL feature ID  : CDLF-00016-00
+```
```
Kononnable, line 171: Didn't we agree to remove some of those fields on the meeting?

docs/rfc/NAF-Documentation_and_RFC_process.md
Kononnable: I'm not sure if we agreed on staying with the `Version` field. I think it was unclear what this field really is.

docs/rfc/NAF-Documentation_and_RFC_process.md
luke-biel: With current amount of RFC discussions I don't think it's doable. All I'd do is read rfc's, probably twice, coz once day before, once same day, to be on par with meeting. Plus meeting, plus time for comments, this grows and grows.

Kononnable (approved):

PR #619, opened 2021-07-02
fix: Object builder returns error trace for gRPC

frondeus:
Tested manually, was blocking #532 . On-demand materializer returns a generic error without details, and details were supposed to go into traces, but because the outcome of on-demand materialization is the stream, traces were with invalid filtered out span. This 
PR fixes that issue, so now a detailed error message can be read from Jaeger.


esavier:
okay, retesting it to be sure

esavier (approved):

PR #622, opened 2021-07-06
docs: add rfc about `/raw` route in QR, QS and QSTS

luke-biel:
Contents of this rfc aren't up for a face to face discussion like others, unless we find something really big in it. It describes what was already added to CDL, with addition of allowing us to disable this feature (proposal is in section #Additional work).

`{{#include ...}}` are mdBook extensions, contents will appear once rendered.


crates/query-router/api.yml
```diff
@@ -22,6 +22,13 @@ paths:
           schema:
             type: string
             example: "15251181-f749-42e0-b4a4-e4b3d90e990d"
+        - name: REPOSITORY_ID
```
Kononnable, line 25: Don't forget to add info somewhere(e.g. message of squash merge commit) about those changes.
Repository id(or name) was added directly to the request or through the `option` subobject?

docs/rfc/CDLF-00017-00-rfc-01.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Glossary
+* `QR` - Query Router
+* `QS` - Query Service
+* `QS-ts` - Query Service - timeseries
+* `gRPC` - Remote procedure call protocol - [website][gRPC]
+* `VM` - Victoria Metrics
+
+# Description
+Raw routes in `QR` and `QS` allow users to bypass CDL api in places where it's lacking.
```
Kononnable, line 20: I wouldn't call it bypassing and api because it's part of the api, semantics.

Kononnable (approved):

docs/rfc/CDLF-00017-00-rfc-01.md
esavier: /shrug

esavier (approved):
i am happy, resolve whatever issues other reviewers have and i can merge this

crates/query-router/api.yml
frondeus: Via `options` subobject

frondeus:
Besides that options subobject - LGTM ;)

crates/query-router/api.yml
luke-biel: This is the other end, requesting data based on REPO_ID header. No body changes.

This change should've been there already, once static routing was introduced, but we don't really keep this OpenAPI spec up to date. I'm not sure if it covers whole QR.

luke-biel:
> i am happy, resolve whatever issues other reviewers have and i can merge this

@esavier I think that's it

crates/query-router/api.yml
frondeus: Oh okay, my mistake!

Also, I wonder whether it should be combined with an RFC, it could be harder to find it or reverse it or whatever, but that's me being a purist. In that particular case, it won't hurt, and reverting it manually is a relatively easy job.

frondeus (approved):

Issue #614, opened 2021-07-02
chore: proper release branching and repinning of workflows


esavier:
related:
also blocks
https://github.com/epiphany-platform/CommonDataLayer/pull/608
- [x] update workflow
- [x] update branching strategy
- [x] make release branch 
- [x] re-release from proper branch
- [x] update documentation sites

PR #623, opened 2021-07-06
chore: update release workflow against new branching strategy

esavier:
workflow update related to changing branches and making proper release branches.
- removed nightlies,
- moved release workflow to work on short and long version taken from the tag
related: #614


.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,44 @@
+name: Release Nightly
+on:
+  push:
+    branches:
+      - 'develop'
+
+
+# its negativbe matching patterns,
```
frondeus, line 8:
```suggestion
# it's negative pattern matching
```

.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Login to registry
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKERHUB_LOGIN }}
+        password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
+    - uses: actions/checkout@v2
+    - name: Build and push crates
+      run: |
+        crates=(data-router command-service query-router query-service query-service-ts schema-registry api edge-registry partial-update-engine object-builder materializer-general materializer-ondemand)
```
frondeus, line 27: The fact that we have to maintain another place with list of crates, it worries me. Maybe we should have `crates.txt` in the root of the repo so we could use this file both in `build.rs` and here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+-rc[0-9]+'
+    branches-ignore:
+      - '*'
+      - '!develop'
+      - '!releases/[0-9]+.[0-9]+'
+
+# its negativbe matching patterns,
```
frondeus, line 12:
```suggestion
# it's negative pattern matching
```

.github/workflows/release-nightly.yml
```diff
@@ -0,0 +1,44 @@
...
+        for crate in "${crates[@]}"
+        do
+          echo "building ${crate}"
+          docker build . --build-arg BIN=$crate -t "epiphanyplatform/cdl-${crate}:lastest"
+          docker tag "epiphanyplatform/cdl-${crate}:latest"
```
frondeus, line 32: Shouldn't we also tag with a timestamp?
Also, you don't have to tag it if the name passed in the previous line is the same

.github/workflows/release-nightly.yml
esavier: Improvements will follow, 
there are a lot of problems with this workflow that i encountered and fixed over night, like for example you can't get both branch name and tag name since only one is provided at a given time. Also since tags are global and not bound to the branch, i do not know how it will behave if we will add tag to some random branch. Hopefully ignored-branches gonna stop it as shown in the first commit, it is not the case.

.github/workflows/release-nightly.yml
esavier: oookay so you want to remove  `docker tag...` completely here?

frondeus: Id say if we want to keep docker image with tag which contains timestamp, for example DD-MM-YYYY, then we can change this line to tag it with DD-MM-YYYY :)

Otherwise we can reverse it and name our image with `DD-MM-YYYY` and then tag it with `latest`, but honestly both solutions should work fine.

So tagging works kinda like symlinks.
You can create text file named "07.07.2021.txt" and then symlink "latest.txt" that points to it, or in reverse.
There is however a difference, when the next day comes, and you want to push another file, when you have a text file "latest.txt" and symlink "07.07.2021.txt" pointing to it, and then you want to create new text file "latest.txt" and symlink "08.07.2021.txt" you don't remove previous latest.txt. Instead, docker is smart enough to know that previous latest.txt is now named only 07.07.2021.txt, so its still available under this name, and latest.txt now points to new build :)

So technically maybe better analogy would be that image has random UUID, and both "latest.txt" and "07.07.2021.txt" points to this UUID...

And when you use docker build with `-t` flag, you actually create this image with some internal UUID, and then tag it :)
Therefore in that case you tagged it twice with the same name.

.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
+name: Release Build
+on:
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+-rc[0-9]+'
+    branches-ignore:
+      - '**'
+      - '!develop'
+      - '!releases/[0-9]+.[0-9]+'
```
luke-biel, line 10: For `branches`, `tags` were `or`ed, which lead to double builds on tag pushes to develop. Are you sure `branches-ignored` don't do the same?

esavier: i was under the impression it is but right now i do not, documentation is made by microsoft people so the quality is what it is do not know how to solve blacklisting the branches beside a job i guess?

.github/workflows/release-nightly.yml
frondeus: Yeah, tagging is global :( Thats why maybe... just maybe we could introduce a bot in the future that would tag for us.

.github/workflows/release-nightly.yml
esavier: i understand how it works, but i do not really understand why do we need daily builds, 
1. noone uses them anyway, its especially visible by looking at the statistics
2. sometimes changes are miniscule yet build still happen
3. repository on dockerhub looks trashy because of this, and images have to sit there unused for 6mths before being removed automatically

frondeus: Okay, that, especially 1. is very good valid point. I guess we could discuss it?

frondeus: @luke-biel , @Kononnable , @smores56 , what do you think? I think we could remove the timestamp, but maybe it is useful to you?

Kononnable: Ok, so the question here is to remove timestamp or daily builds at all? My understanding was that some customers may want to try CDL with features that are not yet finished/unreleased. This is no longer a requirement?
I'm fine with changing them to weekly builds(if it's still a requirement), or removing them at all(if it's not).

frondeus: We could just use one tag like `devel` - then the customer can try CDL with features that are not yet finished/unreleased but at the same time, he can not try CDL from 3 days ago. Which makes the list shorter and cleaner.

Kononnable: I don't think that's a good way.
Let's say I installed CDL locally on minikube on day I, on day II I was playing with it and had to recreate a single service. Woopsie, minikube downloaded the latest version(as requested) and I have CDL running with services from different(maybe incompatible) builds. The only way for me to say if something is a bug or just incompatibility between two different builds is to recreate the whole cluster, insert data again, etc.

Kononnable: And one more thing - you cannot easily recreate the environment in which the customer found a potential bug - the tag points to a different version now.

esavier: 1. i do not know if or why customer would use devel versions and not releases or RC
2. minikube does have cache functionality at least to my udnerstanding, however i am not an expert
3. i think we are still far, far beyond supporting specific images, and if ther eis a bug, it will also be present in either normal release or rc

frondeus: As far as I see no one uses it. I would not use it. If I need that fresh changes I usually just clone the repo and build the docker image manually. Then the issue is solved by the fact that I have a specific commit hash.
Plus for now, clients were patiently waiting for RC versions.

frondeus: I mentioned this only because: We had this functionality and it was important to make sure we are not removing it by accident and otherwise there is unnecessary docker tag

luke-biel: So, a few words from me, i think I missed this in my previous review:
`latest` is usually referred to latest stable release, as a `symlink` to it. If our definition diverges from this, clients will be confused. So we cannot name this `latest`, maybe `devel`. (Although, this opinion is based on observation, this seems to be an unwritten standard.)
If we want to be compliant with this mechanic I mentioned we'd have to change main `publish.yml` to double tag images with latest.
If we don't do this, we'll have `latest` tag pointing to `1.0` for next 6 months, even after we release `1.1` and possibly `1.2`. That's confusing.

As for do we even need nightly builds, I would say, if we are not keeping the history, if nobody uses it, let's delete them and keep only tagged, official ones.

luke-biel: Also, there's a typo `lastest`

esavier: > Also, there's a typo `lastest`
you know, there is last, latest and lastest xD

esavier: your comment does makes sense tbh,
latest was supposed to be tip of the devel branch not the releases, and i would refrain from adding tag latest to any proper release i guess
:rocket: we can change it to devel
:rabbit: we can drop it entirely

esavier: there is a lot of magic to be written if we want to keep latest actually latest, 
git does not allow easy api for that and that what i spent last night on.
for example we have 1.1 and 1.2 and whatever is tagged latest is the one that get's the cherry-pick the least, not actually the latest stable version, and we can have only one latest

smores56: I would just drop it entirely. That avoids a lot of issues with respect to clients not updating version properly, as well as development time on our end.

esavier: yeah, there may be an issue when we will use latest, and then latest will switch for example from 1.X to 2.X

Kononnable: Feels like this is more of a discussion and decision than doing stuff already decided. So we probably should use a standard discussion process on what we want to achieve instead of using PR.
TL;DR: no scope/acceptance criteria on what we want to achieve by this PR when it was created.

If any decisions come out of this it should be announced properly - at least list what was changed in the PR description(in a logical sense, not changed files).

esavier: as discussed in person, i am fixing tagging, and dropping nightlies for now

frondeus (approved):

Issue #581, opened 2021-06-25
Redesign resolve_tree in Edge registry


Kononnable:
Filtering in resolve_tree should utilize mixed strategies of aggregating filters(`AND`, `OR`). Currently, we're always using `AND`. A similar structure to `ViewDefinition::filters` might be used.
 
Issues that should be considered as part of redesign:
- #523 
- #573 
- #574

Acceptance criteria:
* [x] An RFC is written that briefly describes how the algorithm works now and proposes how the algorithm should look.
* [x] Sub-issues mentioned above should be updated with new Acceptance criteria whenever these change
* [x] E2E tests are written (should not pass), which cover common cases.

Kononnable:
Done in #624

PR #624, opened 2021-07-07
docs: rfc edge registry - resolving relation trees

Kononnable:
#581
[Rendered](https://github.com/epiphany-platform/CommonDataLayer/blob/rfc-edge-filtering/docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md)


docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,107 @@
...
+    },
+    InChildObjIds{
+        relation: Identifier,
+        ids: Vec<Uuid>
+    }
+}
+
+struct Relation {
+    relation_id: UUID,
+    identifier: NonZeroU8, // Unique in view definition,
```
frondeus, line 49: Edge Registry does not know about "View" because one Relation can be used in multiple Views. Therefore this `identifier` needs to be removed as well as `type Identifier`

frondeus: Id rather keep separate `Relation` for the view, which could be transformed into ER `Relation`

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,107 @@
...
+
+```rust
+struct TreeQueryResponse {
+    resp: [Uuid] // number of rows * (2 + TreeQuery.relations.len())
+}
+```
+UUID order - base_relation_parent_id, base_relation_child_id, (relation_x_child_id)*
+relation_x - relation with identifier value x
+
+Why single array(Vec) of UUIDs, not Vec<Vec<UUID>> or something similar to distinguish rows - nested vectors means a lot of small allocations which are terrible for cpu cache. Using rust iterators allows us to handle both solutions the same way, so this optimization shouldn't be problematic from code side.
```
frondeus, line 73: There is a high chance we would be more bottlenecked by network IO than by CPU, so it seems to be a premature optimization that makes code more complex. If you disagree with it, perhaps we should create some benchmark comparing both solutions?

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,107 @@
...
+  rpc GetRelation(RelationQuery) returns (RelationList);
+...
+
+message RelationQuery {
+  repeated string relation_id = 1;
+}
+
+```
+
+Parent schema id is not required for fetching relation information. I'm not sure why this field was added there as required. Removing it and allowing to query multiple relations in one request will serve requirements given by object_id. It will also make `ListRelations` obsolete as the same thing should bo possible to achieve by sending empty `RelationQuery`.
```
frondeus, line 103: As far as I understand you want to retrieve information about parent schema id and child schema id not in `ResoleQuery` but in a separate method, is that correct? 
It is worth mentioning it in RFC because for now its seems that parent schema id (and not explicitly mentioned child schema id) is not required at all in the Object Builder which is false.

Also, `ListRelations` or `RelationList`?

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,107 @@
...
+
+Extracting fetching relation metadata from TreeQuery shouldn't cause any performance problems as those two requests can(should) be ran in parallel by OB. It will also allow implementation of [#573](https://github.com/epiphany-platform/CommonDataLayer/issues/573) if we decide to implement it.
```
frondeus, line 105: I do not follow, Should we include relation metadata in TreeQuery or use another method?

docs/rfc/CDLF-0001F-00-rfc-01.md
Kononnable: Identifier here has nothing to do with view(comment is wrong - unique in request, not view definition).

docs/rfc/CDLF-0001F-00-rfc-01.md
Kononnable: This isn't a tradeoff between more network IO and more CPU. Both statistics should have nicer results on the single vec approach.
With the nested Vec approach, the response will be larger(each vec needs to store its own length in standard serialization methods).

If you think using the single vec approach will make code less readable I'm fine with changing it to nested vecs. It will be a bit slower but I doubt it will be noticeable by the user(and current performance of materialization isn't a priority). I don't think running benchmarks is necessary right now - it's hard to measure such things because a lot matters on the types of data you use and hardware setup.

docs/rfc/CDLF-0001F-00-rfc-01.md
Kononnable: Our current proto file contains `ListRelations`.

>As far as I understand you want to retrieve information about parent schema id and child schema id not in `ResoleQuery` but in a separate method, is that correct?
Yes, in `GetRelation`(this section describes this method). OB can request 'relation definition' based on view definition as parallel request with `resolve_tree`. This shouldn't affect performance and keeps different kinds of information separated(schema vs data) plus allows caching.

docs/rfc/CDLF-0001F-00-rfc-01.md
Kononnable: TreeQuery - response = data
GetRelation - response = schema(in general meaning, not cdl schema)

I think this should make sense now(with the previous comment) - if not mention me on chat. It will be easier to get to what's the unknown here and I'll try to provide a better explanation here.

docs/rfc/CDLF-0001F-00-rfc-01.md
frondeus: Okay, makes more sense, also I'd like to distinguish that this is `RelationRequest` to distinguish it from `Relation` stored in ER, and returned via `GetRelation`

docs/rfc/CDLF-0001F-00-rfc-01.md
frondeus: Current proto file might contain ListRelations but in this RFC I can see RelationList and I wonder if this is the same message.

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,107 @@
...
+### Processing TreeQuery request
+
+- Resolving relation tree should be handled by using `inner join` strategy. This may be configurable in the future.
+- Filtering based on object_ids should be performed.
+- It might be a good idea to let sql engine to do almost all the work for us and just generate sql query and collect the results. Efficient joining and filtering data is one of the main goals of rdbms and it doesn't require as much network chatter as doing it on our own. This is just a proposal - implementator is free to choose which way this should be handled.
+
+
+### <a name="getting_relation"></a>Getting relation information
+
+```proto
```
frondeus, line 91: Perhaps it would be the best to introduce diff, whats changed?

Right now RFC assumes that reviewer or reader has open `edge_registry.proto` on the side, which might look much much differently in the future. It means that this RFC could be not readable in few months (or years) and also it is not self-contained (of course there is no strict requirement to always make RFC self contained).

docs/rfc/CDLF-0001F-00-rfc-01.md
frondeus: Okay, nevermind, my fault at reading this sentence.

docs/rfc/CDLF-0001F-00-rfc-01.md
frondeus: Okay now I understand, Parent schema id is currently required in RelationQuery message and this RFC informs that it is not needed anymore.

esavier:
i am sorry but where is RFC?
this is like change log, i have no idea what is happening

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,150 @@
...
+    child_schema_id: Uuid,
+}
+```
+
+## Proposed changes
+
+### TreeQuery request
+
+```rust 
+type Identifier = u8; // id from TreeQuery->Relation, 0 for base_relation
```
luke-biel, line 47: Wouldn't it be more explicit to use Option<NonZeroU8>?

frondeus: In Object Builder we use Option<NonZeroU8>.

frondeus:
Please rename RFC filename to for example: `0017_Edge_registry_resolving_relation_trees_01.md`

docs/rfc/CDLF-0001F-00-rfc-01.md
Kononnable: Use Option with what type parameter, what would none/some represent. Tbh. I'm not quite grasping what you're suggesting.

The identifier represents unique schema usage in view - for most cases, it could be represented as schema_id. However, due to a fact that we allow schema to be used multiple times in a single view(join to the same schema multiple times on different join conditions, tree-like structures, etc.), we have to be able to identify to which AST element we're talking about.

Kononnable:
I'm not fixing conflicts for now as those two files require manual conflict resolution almost on every change and are quite hot right now(with all other RFCs and changes to the RFC process). Will fix that after discussion when the document is ready for final approval.

@esavier Please check if your comment still applies to this RFC(after RFC changes and changes to the RFC process). If so please specify what exactly you think needs to be changed(constructive criticism).

docs/rfc/CDLF-0001F-00-rfc-01.md
luke-biel: Yeah, sorry, I meant `Option<NonZeroU8>` to indicate that 0 is an unique value. Rust under the hood can simplify union tag and use u8 same way as we do right now.

Kononnable: Do you mean treat reference to base_relation as None, and references to relations in `relations` property with numeric values?
For me, it doesn't really matter(implementation detail). I prefer giving all of them a number, but it's strictly a personal opinion. If GRPC recognizes `Option<NonZeroU8>` properly(doesn't add an unnecessary place for full u8) I'm fine with both approaches. If not we can always write a converted.

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+
+## Glossary
+- Relation - link between objects of two schemas. Relation represents information on a type level, it doesn't connect two specific objects, it simply states that objects of the following schemas may be related to each other
+- Relation graph - information about multiple related objects, single relation graph can present data from multiple relations, currently presented in form of a tree, proposal changes representation form to table
+- TreeQuery - part of Edge registry API responsible for generating relation graphs for provided recipe and filtering criteria
+
+## Motivation
+Our current process on resolving relation graphs is far from ideal - e.g. it allows querying from only one side of the relation, doesn't support complex filtering. Current implementation doesn't cover all of the features required by the materialization phase II(e.g. complex filtering).
```
esavier, line 19: > ... of the relation, doesn't support  ...

isn't there something missing, `and` for example?

> required by the materialization phase II

why is it there? its arbitrary group, just write its required by planned features

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -0,0 +1,160 @@
...
+
+```rust
+struct TreeQueryResponse {
+    resp: [Uuid] // number of rows * (2 + TreeQuery.relations.len())
+}
+```
+UUID order - base_relation_parent_id, base_relation_child_id, (relation_x_child_id)*
+relation_x - relation with identifier value x
+
+Why single array(Vec) of UUIDs, not Vec<Vec<UUID>> or something similar to distinguish rows - nested vectors means a lot of small allocations which are terrible for cpu cache. Using rust iterators allows us to handle both solutions the same way, so this optimization shouldn't be problematic from code side.
```
esavier, line 118: this is implementation specific, there are allocators and/or structures that can handle that 

additionally writing a sentence starting from why is far from correct

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -0,0 +1,160 @@
...
+- no relation information(only edges) - in old format same information about relations was sent multiple times, more info in [Getting relation information](#getting_relation)
+
+It might seem that this format introduces duplication(some ids will be sent multiple times), however that shouldn't be a problem. This is the ideal case for usage of lossless algorithms on protocol level(if needed). Similar solutions(table results) are widely used with rdbms and I don't recall any performance issues about sending the data.
+
+### Processing TreeQuery request
+
+- Resolving relation tree should be handled by using `inner join` strategy. This may be configurable in the future.
+- Filtering based on object_ids should be performed.
+- It might be a good idea to let sql engine to do almost all the work for us and just generate sql query and collect the results. Efficient joining and filtering data is one of the main goals of rdbms and it doesn't require as much network chatter as doing it on our own. This is just a proposal - implementator is free to choose which way this should be handled.
```
esavier, line 131: what if sql engine does not have such feature or it is not sql engine we are getting data from?

esavier:
> i am sorry but where is RFC?
> this is like change log, i have no idea what is happening

...again, as said before, this is like a change log or planning suggestion, i do not know what problem are we trying to solve, there is little to no explanation of what is happening. There are no requirements mentioned, there are no assumptions, nor anything else that would help understand the standpoint of the writer else.

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
Kononnable: There is a comma, I think it's more appropriate as this list is not complete(e.g.)

As for using `materialization phase II` - I don't see a problem in being specific here. This term is already used in multiple GH issues, it's even defined in one of them.

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
Kononnable: All of the mentioned features are part of early SQL standards, so they exist in all popular SQL Engines.
If we want to migrate Edge registry storage from Postgres to something else we shall decide on ups and downs then - same as with other CDL components.

Kononnable:
The decision from the meeting - proposed idea accepted. There will be some changes to RFC text(not the idea) done by @esavier.

frondeus:
looks good to me, please resolve conflicts and we can merge it :)

esavier (approved):

PR #625, opened 2021-07-07
chore: Rename Array into SubObject

frondeus:

PR #626, opened 2021-07-07
Rename Array to SubObject for FieldDefinition

smores56:
Included conversion from Array and logging warnings.


frondeus:
Duplicated with #625 - can be closed.

PR #631, opened 2021-07-08
docs: RFC - Simplify Schema Definitions

smores56:
Scheduled meeting is needed


docs/rfc/CDLF-00018-00-rfc-01.md
```diff
@@ -0,0 +1,161 @@
...
+ObjectFieldDefinition:
+```
+{
+   name: String,
+   type: ChildType,
+   optional: Bool,
+}
+```
+
+SchemaDefinition:
```
Kononnable, line 56: This is the definition of schema or field in the schema? From provided types, it looks like there can be only one SchemaDefinition in Schema - so definition may consist of one Scalar field without a name.

docs/rfc/CDLF-00018-00-rfc-01.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+ScalarType:
+```
+Null | Bool | String | Number | Any // Any represents arbitrary JSON data
+```
+
+ChildType:
+```
+Scalar(ScalarType) | ChildSchema(UUID) // ID of other Schema
```
Kononnable, line 44: Child schema is used only on schema creation(resolve on creation) or on every request(resolve on fetch)?
This will change behavior if we decide to allow schema mutability in the future. For now, it's mostly about doing a little more work on fetch vs not remembering that child schema exists as a separate thing(inlining).

docs/rfc/CDLF-00018-00-rfc-01.md
```diff
@@ -0,0 +1,161 @@
...
+    id: UUID,
+    name: String,
+    definition: SchemaDefinition,
+    ... // other configuration metadata
+}
+```
+
+ScalarType:
+```
+Null | Bool | String | Number | Any // Any represents arbitrary JSON data
```
Kononnable, line 39: I'm not sure if null is reasonable here. Something or null - sure, but null on its own. It would allow for situations where all objects should have null in one property(no real information here).

frondeus:
You should probably add it also to the Feature List

docs/rfc/CDLF-00018-00-rfc-01.md
frondeus: So you would like to have some kind of Option<ScalarType> as a variant of ScalarType?

docs/rfc/CDLF-00018-00-rfc-01.md
```diff
@@ -0,0 +1,161 @@
...
+
+In the case that some JSON data is too complicated to be represented by the proposed system, or a user would like
+to opt out of the optimizations provided by properly-typed schemas, the `Any` scalar variant will accept any valid
+JSON data.
+
+### Use in Materialization
+
+With this system, all fields in a schema definition have deterministic types. Given that, any view that references
+fields from an `Object` schema definition can know the field's type. Either the type is a potentially nullable scalar,
+which can be represented by a SQL type, or it is a more complicated type (or an `Any` scalar), and it falls back to
```
frondeus, line 126: How about other databases than SQL? Is nullable scalar anyhow affecting it?
Should we keep, perhaps in this RFC a table for mapping between our Type and SQL column?

docs/rfc/CDLF-00018-00-rfc-01.md
```diff
@@ -0,0 +1,161 @@
...
+a JSON type. That way, all materialized data can have its fields stored in its respective types.
+
+### Schema Composition
+
+There is a desire from CDL users to define data that is related in definition to other schemas, whether that be
+via inheritance or composition. As composition is generally less problematic and simpler to implement than
+inheritance, it can be implemented using the approach outlined in this proposal. When creating a definition for
+a schema, users can refer to another existing schema's definition by linking to that schema in an `Array` or an
+`Object`'s field(s).
+
```
frondeus, line 136: I am not sure whether we can simply discard Schema Inheritance

For now, with the only composition we allow such a definition:
Schema A: 
```
{
   foo: String,
   bar: Number
}
```
Schema B:
```
{
   baz: String,
   a: { //Object type with reference to SchemaA
      foo: String,
      bar: Number
   }
 }
 ```
 
 This works in some cases but of course discards definitions like:
 
 SchemaC
 ```
 {
  baz: String,
  //Definition is extended by using a reference to SchemaA
  foo: String,
  bar: Number
 }
 ```
 Of course, it means there is a possibility for name collision - what id SchemaC defines `foo` too? Should it be shadowed?
 It is more complex, but also flexible. Are we 100% sure that limiting won't be a problem?

File: docs/rfc/CDLF-00018-00-rfc-01.md (hunk 2 above)
Comment (frondeus) [reply docs/rfc/CDLF-00018-00-rfc-01.md:44]: If we decide to keep child schema ID in our system (and not expanding it in Schema Registry on creation) it would require Object Builder to expand it by multiplying the number of calls to Schema Registry.

We could, however, create a gRPC method in Schema Registry which would resolve it on the fly. Then for the other services `ChildSchema` would be unreachable because these would always work on fully expanded schemas.

Also - I guess we should prevent making cycles, am I right?

File: docs/rfc/CDLF-00018-00-rfc-01.md (hunk 1 above)
Comment (frondeus) [reply docs/rfc/CDLF-00018-00-rfc-01.md:56]: Not really, if you need JSON object you can always set SchemaDefinition to Object type variant.
This RFC would allow data to be scalars, while currently we always expect it to be an object.

To be more clear, now you are not allowed to send:
```
{
  "object_id": "...",
  "schema_id": "...",
  "version": "...",
  "options": {...},
  "data": 5.0
  }
  ```
  Because `data` has to be an object.
  With `SchemaDefinition` it doesn't.
  
  Of course it requires extra thinking - if data send to the CDL is `Scalar` or an `Array`, what type should we use to store materialized view? 
  
  With `Object` it is very simple, we create SQL table, where each column represents field in the object.
  
  How are we going to store `Scalar`? By introducing default column name `value`? 
   
  And what about `Array`? Should we force that an array has to be defined with known static size and put each element in separate column? Or use SQL array field? Or use extra table with foreign key? 
  
  It is complex and I only asked questions about SQL. Whenever we introduce new repository, the complexity will only grow.
  
  Also I am not sure how to handle Victoria Metrics here.

docs/rfc/CDLF-00018-00-rfc-01.md
Kononnable: The only scenario (that I see), where information about child schema may come in handy is administration - retrieving schema definition manually by the administrator or modifying schema definition. For all other purposes, I think the schema registry can respond with inlined version.

As for cycles:
- full cycles are mistakes for sure - creating them should be checked in the code for sure
- the possibility of cycles isn't a technical mistake - the user may want to create even a self-referencing schema if reference field is optional(or nullable) - any tree structure is defined that way. However, we should decide/get info from potential customers if we want/need to support such cases. It will be harder to implement as we won't be able to 'inline' schema - validation and similar features based on schema definition will get much more complicated.

docs/rfc/CDLF-00018-00-rfc-01.md
Kononnable: I mean types like `Bool | Null`(or `Option<Bool>`) are fine, `Null` on its own shouldn't be allowed as field type because it will force each object to contain property X, but its value will always be null. If schema guarantees field value to be specific there is no real advantage in storing this value at all.

Sidenote: if/when we introduce null type there will be some decisions to be made about those types - e.g. standard SQL question: does any of those return true value? 
- `NULL = NULL`
- ` NULL in (1, 2, 3)`
- `1 in (NULL)`

docs/rfc/CDLF-00018-00-rfc-01.md
Kononnable: This can be done in composition(composition can be flat, doesn't necessarily require another subobject), however, it's not described in this RFC.

docs/rfc/CDLF-00018-00-rfc-01.md
smores56: Yeah, I was thinking the same thing. I'll drop normal null and only maintain nullability.

docs/rfc/CDLF-00018-00-rfc-01.md
smores56: For anything that we can't determine the type of, we can always store raw JSON. However, even if a schema just stores booleans or arrays, a view referring to a scalar will provide a structure like an object, where the whole schema is now a row, instead of a field from the schema.

docs/rfc/CDLF-00018-00-rfc-01.md
smores56: If we introduce schema mutation in the future, I would again prefer something like versions. For now, without mutation, fetching on every "request" (where the SR combines Schemas together on request) is the "correct" way to do it. I'd only want to change that if we realize that this approach is way too inefficient, which I don't believe it is.

smores56: > The only scenario (that I see), where information about child schema may come in handy is administration - retrieving schema definition manually by the administrator or modifying schema definition. For all other purposes, I think the schema registry can respond with inlined version.
> 
> As for cycles:
> 
>     * full cycles are mistakes for sure - creating them should be checked in the code for sure
> 
>     * the possibility of cycles isn't a technical mistake - the user may want to create even a self-referencing schema if reference field is optional(or nullable) - any tree structure is defined that way. However, we should decide/get info from potential customers if we want/need to support such cases. It will be harder to implement as we won't be able to 'inline' schema - validation and similar features based on schema definition will get much more complicated.

Yeah, children probably won't need it. But it is mainly useful for defining relationships from parents to children to allow for schema re-use.

docs/rfc/CDLF-00018-00-rfc-01.md
smores56: Yeah, I have now changed Scalars to `Boolean | String | Number`, and SchemaDefinitions now have the Scalar variant defined as `Scalar { type: ScalarType, optional: Bool }`.

docs/rfc/CDLF-00018-00-rfc-01.md
smores56: Yes, technically composition can allow for these things, but I would prefer to avoid allowing users to do this unless they really want to. I have experienced terrible data structure inheritance many places.

However, it's not up to me. I'll add a description for inheritance as an alternative option, and why I'm weary of adding it.

docs/rfc/CDLF-00018-00-rfc-01.md
smores56: Sure, that can be added.

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,193 @@
...
+In the case that some JSON data is too complicated to be represented by the proposed system, or a user would like
+to opt out of the optimizations provided by properly-typed schemas, the `Any` scalar variant will accept any valid
+JSON data.
+
+### Use in Materialization
+
+With this system, all fields in a schema definition have deterministic types. Given that, any view that references
+fields from an `Object` schema definition can know the field's type. Either the type is a potentially nullable scalar,
+which can be represented by a SQL type, or it is a more complicated type (or an `Any` scalar), and it falls back to
+a JSON type. That way, all materialized data can have its fields stored in its respective types.
```
Kononnable, line 138: There is a small hiccup when it comes to knowing exact SQL types. They don't map 1:1.
Examples:
- String - mentioned later; `varchar`(with defined max length) or `text`
- Number - multiple sizes and precisions, the difference between arbitrary precision numbers and floating-point types

Anyway, this problem doesn't have to be solved here. Adding proper schema types will add value to CDL even without solving all materialization problems related to data types.

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Further Considerations
+
+### Impact on Other Teams/Clients
+
+Though validation is not currently being done on incoming data and therefore schema definitions don't matter
+much at the moment, this will change not only all currently stored schema definitions but also all future schema
+definition insertions. There will potentially need to be a one-time migration if anyone wants to keep their
+schema definition. Likely, the volume of such request for conversion will be low enough to be doable manually.
+
```
Kononnable, line 178: Do we treat it as a breaking change?

docs/rfc/CDLF-0001F-00-rfc-01.md
Kononnable: For string this shouldn't be much of a problem - we may make a choice(an educated guess) and there will be no logical(functional) problems - only performance may differ.
However, with numbers, it's a different matter. There is no clear winner in arbitrary precision vs floating-point and in the future, we should probably support both types(maybe even in schema definitions as we may store data in command service differently).

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,193 @@
...
+} |
+Object {
+    fields: List<ObjectFieldDefinition>,
+} |
+Array {
+    item_type: ChildType,
+    optional: Bool,
+}
+```
+
```
Kononnable, line 76: Just to make sure: do we want to allow definitions containing only a single field, without a field name? I understand this as a feature but I'm not sure about the costs of implementing it - this change will have to be handled in multiple places.

Also, I'm not sure about Child Type in the array - what will happen if someone decides to put a child schema there espessially with different merge values.

docs/rfc/CDLF-0001F-00-rfc-01.md
smores56: Yeah, something we can support differently now that we aren't just using JSON Schema. Maybe having an Integer and a Float would work?

docs/rfc/CDLF-0001F-00-rfc-01.md
smores56: Yes, this would be technically a breaking change. We could put it under the rug for now by just having people default to an `Any` typed-schema, and then have them manually convert (not breaking), or force them to try to convert by hand (definitely breaking). Either way, they should have new schemas created this way, which is a breaking change.

docs/rfc/CDLF-0001F-00-rfc-01.md
smores56: I'm fine with supporting only schemas that are top-level objects, and can have whatever fields. Does anyone oppose simply using objects instead of supporting any arbitrary JSON? Note: this would be a breaking change if anyone has scalar JSON schemas defined (unlikely).

smores56: It would be simpler to just use objects for a lot of reasons.

frondeus: I think this is a good topic for scheduled meeting discussion.

docs/rfc/CDLF-0001F-00-rfc-01.md
frondeus: We need some kind of table in documentation that maps our JSON type to SQL type

frondeus:
Please rename RFC filename to for example: 
`0016_Simplify_schema_definitions_01.md`

docs/rfc/CDLF-0001F-00-rfc-01.md
smores56: This is appropriate for docs, but maybe not this RFC? We can discuss that in tomorrow's meeting.

crates/materializer-ondemand/src/lib.rs
```diff
@@ -42,7 +42,7 @@ impl OnDemandMaterializer for MaterializerImpl {
             .map(|(k, v)| (k, into_object_builder_schemas(v)))
             .collect();
 
-        let stream = rpc::object_builder::connect(self.object_builder_addr.as_str())
+        let stream = rpc::object_builder::connect(self.object_builder_addr.clone())
```
frondeus, line 45: This is not a part of RFC, should be included in separate branch and PR

frondeus (changes requested):
Probably a mistake during commiting, rejecting last changes.

docs/rfc/0001F_Simplify_Schema_Definitions_01.md
```diff
@@ -0,0 +1,194 @@
+# Front Matter
```
frondeus, line 1: I see you used 0001F as an ordering number. I don't have strong feelings about that one but maybe we should use a different one?
In the RFC process we said: "RFC ordering number and feature ID are not the same concepts".

Probably we could simply use a decimal integer as an ordering number. See: https://github.com/epiphany-platform/CommonDataLayer/issues/645 - I proposed ordering numbers for existing RFCs.

If you disagree, then fine, we can discuss another approach but Id prefer to be consistent.

esavier:
please resolve outdated threads

esavier (dismissed):

docs/rfc/0001F_Simplify_Schema_Definitions_01.md
```diff
@@ -177,6 +152,9 @@ much at the moment, this will change not only all currently stored schema defini
 definition insertions. There will potentially need to be a one-time migration if anyone wants to keep their
 schema definition. Likely, the volume of such request for conversion will be low enough to be doable manually.
 
+However, we plan to explicitly document to our end users that we do not plan to run the migrations for them, as
+it we do not currently have any users relying on this feature and we
```
Kononnable, line 156: Looks like someone ate the end of this sentence.

docs/rfc/0001F_Simplify_Schema_Definitions_01.md
smores56: I changed it to use an integer for the RFC number and the next hex value for the feature number.

docs/rfc/0001F_Simplify_Schema_Definitions_01.md
smores56: Fixed!

frondeus (approved):

esavier (approved):

Issue #404, opened 2021-04-27
feat: Transport Headers passthrough


esavier:
We need to implement a passthrough mechanism for transport headers, like kafka and RabbitMQ.
- this have to be generic enough. Lack of headers should not create issues, and its presence should be handled
- consider putting it behind config flag

luke-biel:
Should we store these headers in db and emit them with notifications? Do they affect materialization? Will our clients write their own `CS` that would handle headers?

Kononnable:
@esavier What's the use case?

frondeus:
Extracted RFC into #653

PR #635, opened 2021-07-09
docs: Transport headers passthrough RFC

frondeus:
RFC for #404

- [x] Rename RFC filename to `0018_Transport_headers_passthrough_01.md`
- [x] Write notes after the scheduled meeting


docs/rfc/template.md
```diff
@@ -0,0 +1,21 @@
...
+## Glossary
+## Formats
+## Introduction
+## Proposed solution
+## Further Considerations
+### Impact on other teams
+### Scalability
+### Testing
+
+[gRPC]: https://grpc.io/
```
frondeus, line 20:
```suggestion
```

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,119 @@
...
+Then headers can be retrieved by using `BorrowedMessage` and `BorrowedHeaders`. Unfortunately headers are not stored as a map.
+Instead, one has to take `.count()` and then retrieve all headers manually. 
+
+It might be wise to reuse `KafkaConsumerHeaders` from `tracing_utils`.
+
+### AMQP
+AMQP supports message headers in field table format: [spec][AMQP].
+
+#### in Rust
+Rust implementation, lapin, supports headers by providing [`with_headers`][Lapin-Set] and [`headers`][Lapin-Get] methods in `AMQPProperties`:
```
frondeus, line 57:
```suggestion
Rust implementation, lapin, supports headers by providing [`with_headers`][Lapin-Set] and [`headers`][Lapin-Get] methods in `AMQPProperties`.
```

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,123 @@
...
+## Header conflicts
+
+Based on the fact that CDL is using headers for Open Telemetry support, and in the future, it might be possible to use more headers for other purposes.
+
+Therefore implementation should either:
+* allow for header shadowing: new header has higher priority, and consequently, it overrides the old one,
+* keep only the first header: old header has higher priority, and therefore application ignores new insertions by always keeping the original value,
+* return an error when such a situation occurs: header with the name `XXX` is forbidden in the CDL and can be used only internally by CDL
+
+As an outcome of this RFC implementation, one must provide documentation enlisting all headers (including one used for Open Telemetry) used internally in CDL to inform the user what may generate such a conflict.
```
Kononnable, line 82: Can we add a prefix to CDL headers? 
As for open telemetry - this might be a specific case(different from how we generally handle headers). Reusing header values here will allow for better observability of clients' systems (won't hide CDL internals).

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Configurability
+
+Header passthrough, as well as conflict resolution, should be configurable.
+
+Preferably conflict resolution strategy should be configurable on a global (configuration of the service) level or/and in the local schema or object level (TODO: discuss).
+
+Header passthrough should be available under the feature flag, which could be by default disabled to prevent any breaking changes; however, whenever CDL reaches a new MAJOR version, it might be wise to reconsider enabling it by default.
```
Kononnable, line 90: I'm not sure that this will introduce a breaking change. We're considering adding a behavior. Not many things can break here - header conflict resolution may 'decide' if this new behavior can(technically) break some code. Then it's up to us(or PO) to decide how likely this can break user code - if it's highly unlikely this may not be classified as a breaking change.

I would advise against classifying something as breaking change prematurely.

docs/rfc/CDLF-0001F-00-rfc-01.md
frondeus: > Can we add a prefix to CDL headers?
Yes we could.

> As for open telemetry - this might be a specific case(different from how we generally handle headers). 
Agree

> Reusing header values here will allow for better observability of clients' systems (won't hide CDL internals).
Exactly

esavier (dismissed):

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,123 @@
...
+* keep only the first header: old header has higher priority, and therefore application ignores new insertions by always keeping the original value,
+* return an error when such a situation occurs: header with the name `XXX` is forbidden in the CDL and can be used only internally by CDL
+
+As an outcome of this RFC implementation, one must provide documentation enlisting all headers (including one used for Open Telemetry) used internally in CDL to inform the user what may generate such a conflict.
+
+## Configurability
+
+Header passthrough, as well as conflict resolution, should be configurable.
+
+Preferably conflict resolution strategy should be configurable on a global (configuration of the service) level or/and in the local schema or object level (TODO: discuss).
```
frondeus, line 88: Lets discuss it here

Kononnable: If we're going to make it configurable on the object level we need to introduce a mechanism for that(we have no similar feature right now). For now, I would do no such thing as the problem can be easily solved globally.

If we can introduce a prefix for our usage of headers I think we can go with the simplest(and popular in such cases)  approach - shadowing. With prefix name collisions are not much of a problem, we can introduce an optional configurable prefix if we want to be fully bullteprof. This will also mean much less work(now and maintenance later) as only one(and simplest to implement) strategy is used.

Apart from the prefix, I don't see what else can/needs to be made configurable here. The whole feature should not be hidden under a feature switch.

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,123 @@
...
+
+Header passthrough should be available under the feature flag, which could be by default disabled to prevent any breaking changes; however, whenever CDL reaches a new MAJOR version, it might be wise to reconsider enabling it by default.
+
+## Alternatives
```
luke-biel, line 92: I'm considering scenario where these headers would be passed from DR to CS as part of the payload, and later on converted back to corresponding headers during creation of notifications. Benefit of this approach is that such headers are fully transparent to the system as external metadata and are independent to messaging protocol.

```json
{
  "objectId": "...",
  "schemaId": "...",
  "timestamp": 0,
  "metadata": {
    "headers": []
  },
  "payload": {}
}
```

docs/rfc/CDLF-0001F-00-rfc-01.md
```diff
@@ -0,0 +1,123 @@
...
+Rust implementation, lapin, supports headers by providing [`with_headers`][Lapin-Set] and [`headers`][Lapin-Get] methods in `AMQPProperties`.
+
+The implementation of this feature should be quite similar to one described in Kafka section, although fortunately, lapin uses `BTreeMap` as a backend for `FieldTable`.
+
+### gRPC
+gRPC is based on HTTP2; therefore, it is possible to add custom headers.
+Based on [gRPC protocol][gRPC-Protocol], it would be wise to use `Custom-Metadata` for that purpose.
+It is stored in a separate header, so there is no chance for collision between the header set by the user and what gRPC uses internally.
+
+#### in Rust
```
luke-biel, line 66: Are there any headers that we would like to skip in case of gRPC? It's common that http clients push a lot of boilerplate in headers, bloomRPC for example supplies: 
```json
{"te": "trailers", "content-type": "application/grpc", "user-agent": "grpc-node/1.24.7 grpc-c/8.0.0 (osx; chttp2; ganges)", "grpc-accept-encoding": "identity,deflate,gzip", "accept-encoding": "identity,gzip"}
```

luke-biel: Maybe I have to rephrase what I wrote: is there a client or specific use case we can base this on?

Kononnable: We probably should decide how we should handle grpc in general.
With message queues we've got notification as a response to a request. With grpc we send a response and notification as an request to another endpoint. Sending back headers in a response is pointless as client already have access to request message(and its headers). Sending headers to notification endpoint may have some value, but we'll probably override some protocol related headers(e.g. user-agent), however we may pass some of protocol related headers without overriding them - e.g. we can override `user-agent` but keep `accept-encoding` intact(our client may not set this header on its own). It may be difficult to tell which headers were overriten by us, which were just passed. On the bright side those 'protocol headers' are mostly not tied anyhow to business logic of an aplication(so this shouldn't be much of a problem)

frondeus:
TODO: Rename RFC filename to `0018_Transport_headers_passthrough_01.md`

docs/rfc/0018_Transport_headers_passthrough_01.md
```diff
@@ -0,0 +1,127 @@
...
+* keep only the first header: old header has higher priority, and therefore application ignores new insertions by always keeping the original value,
+* return an error when such a situation occurs: header with the name `XXX` is forbidden in the CDL and can be used only internally by CDL
+
+As an outcome of this RFC implementation, one must provide documentation enlisting all headers (including one used for Open Telemetry) used internally in CDL to inform the user what may generate such a conflict.
+
+## Configurability
+
+Header passthrough, as well as conflict resolution, should be configurable.
+
+Preferably conflict resolution strategy should be configurable on a global (configuration of the service) level or/and in the local schema or object level (TODO: discuss).
```
Kononnable, line 88: Do we leave this TODO? This RFC ended up being research only(different requirements) so it may be fine(doesn't have to solve everything).

docs/rfc/template.md
```diff
@@ -0,0 +1,19 @@
+# Front Matter
```
Kononnable, line 1: Leftovers or just something that should be committed, but not related to the RFC?

docs/rfc/0018_Transport_headers_passthrough_01.md
```diff
@@ -0,0 +1,127 @@
...
+
+Header passthrough should be available under the feature flag, which could be by default disabled to prevent any breaking changes; however, whenever CDL reaches a new MAJOR version, it might be wise to reconsider enabling it by default.
```
Kononnable, line 90: Adding new optional field/behavior that won't break existing code isn't a breaking change(afaik).

frondeus: If we introduce headers passthrough with header conflict resolution then it can be considered a breaking change, it can, after all, result in an error (conflict between headers) that was not present before.

Kononnable (approved):

Issue #405, opened 2021-04-27
Materialization Phase II


Kononnable:
Features: 
-  complex filters
- views based on multiple schemas(joins)
- computed values
- basic data types in materialized tables(postgres) - or rather building blocks for future feature

Field names may differ, however same information should be included in the same structure.
Note: Some fields might be optional on view creation(like field types), SR should populate such values if not provided.
ViewDefinition:
```rust
struct ViewDefinition {
    base_schema: UUID,
    fields: HashMap<String, Field>,
    filters: Option<Filter>,
    schemas: [Relation],
    materializer: MaterializerOptions, // materializer related options(not described)
}

type Identifier = u8; // id from viewDefinition->Relation, 0 for base_schema

enum Field {
    Simple {
        field_name: String,
        field_type: FieldType,
    },
    Computed {
        computation: Computation,
        outputFieldName: String,
        field_type: FieldType,
    },
    Array {
        base: Identifier,
        fields: HashMap<String, Field>,
    },
}

// To be extended later - new enum variants and data(e.g. db specific types for materializer to create better tables)
enum FieldType {
        String,
        Numeric,
        JSON,
}

struct Computation {
    operator: ComputationOperator,
    lhs: Computation,
    rhs: Option<Computation>,
}

// To be extended later
enum ComputationOperator<T: FieldType> {
    RawValue {
        value: T,
    },
    // base schema - None(or same as base schema) if Computation operator used in ViewDefinition->fields, FieldPath - path to field using dot notation
    FieldValue {
       //  None(or same as base schema) if Computation operator used in ViewDefinition->fields,
        base_schema: Option<UUID>,
       //  path to field using dot notation
        field_path: string,
    },
    Equals,
}

enum Filter {
    SimpleFilter,
    ComplexFilter,
}

struct SimpleFilter {
    operator: FilterOperator,
    lhs: FilterValue,
    rhs: Option<FilterValue>,
}

enum FilterValue {
    SchemaField {
        schama_id: Identifier,
        field_path: String,
    }, // Path to field defined with dot notation
    ViewPath {
        field_path: String,
    }, // Path to field defined in ViewDefinition->Fields, using dot notation,
    RawValue {
        value: String,
    },
    Computed {
        computation: Computation,
    },
}

// To be extended later
enum FilterOperator {
    Equals,
}

struct ComplexFilter {
    operator: BitOperator,
    operands: [Filter],
}

// To be extended later
enum BitOperator {
    And,
[...]

PR #637, opened 2021-07-13
chore: e2e materialization tests

### LINKED ISSUE #618: Define and add acceptance tests for features introduced in materialization phase II

Kononnable:
closes #618

E2E tests for functionalities covered by materialization phase II (#405). 
Some tests are currently placeholders(no code) as some part of the API is under redesign(#624, #631)


e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
...
+        let field_c = row.fields.get("field_c").unwrap().0.as_bool().unwrap();
+        assert_eq!(field_a, "1");
+        assert!(field_b);
+        assert!(!field_c);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn should_compute_field_from_another_object() -> Result<()> {
```
luke-biel, line 391:
```suggestion
    async fn should_compute_field_when_more_than_one_objects_in_relations() -> Result<()> {
```

I don't see a place where you'd take value from object b for computation.

e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
...
+            )
+            .await?;
+
+            sleep(Duration::from_secs(1)).await; // async insert
+
+            let view_data = materialize_view(view, &[schema_a]).await?;
+            assert_eq!(view_data.rows.len(), 2);
+            for row in view_data.rows {
+                let field_a = row.fields.get("field_a").unwrap().0.as_u64().unwrap();
+                assert!(field_a == 1 || field_a == 2);
```
luke-biel, line 673: I would prefer more explicit asserts as tests are a form of documentation. If you have problems with order of values, use `.sort(&mut self)`.

This test will accept result `[{field_a: 1}, {field_a: 1}]`.

e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
...
+            sleep(Duration::from_secs(1)).await; // async insert
+
+            let view_data = materialize_view(view, &[schema_a]).await?;
+            assert_eq!(view_data.rows.len(), 2);
+            for row in view_data.rows {
+                let field_a = row.fields.get("field_a").unwrap().0.as_u64().unwrap();
+                let field_b = row.fields.get("field_b").unwrap().0.as_u64().unwrap();
+                assert_eq!(field_a, field_b);
+                assert!(field_a == 1 || field_a == 2);
+            }
```
luke-biel, line 787: Same as above - please convert to explicit asserts instead of loop.

e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
+use crate::{api::*, *};
```
luke-biel, line 1: Notes about readability:

I'm a fan of using tests as a way to document processes, and any file with over 1000 lines won't be really read by developers - sifting though it is difficult.

I would love to see some adjustments that would help improving this readability, my ideas are:
* put submodules in separate files
* use macros to reduce boilerplate regarding building Filter or Compute fields

I understand that there are e2e that will supersede this soon, but we don't yet know when.

e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
...
+            let field_b = row.fields.get("field_b").unwrap().0.clone();
+            let field_c = (field_b["field_c"]).as_u64().unwrap();
+            assert_eq!(field_c, 1);
+
+            Ok(())
+        }
+
+        #[tokio::test]
+        #[ignore = "todo"]
+        async fn should_allow_filtering_using_field_not_materialized_in_view() -> Result<()> {
```
luke-biel, line 874: This behavior has to be documented for end user.

luke-biel (changes requested):

e2e/src/on_demand_materializer.rs
frondeus: I vote for option 1

e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
...
+                relations: vec![],
+                search_for: SearchFor::Children,
+            }],
+            None,
+        )
+        .await?;
+
+        let object_id_a = Uuid::new_v4();
+        let object_id_b = Uuid::new_v4();
+        insert_message(object_id_a, schema_a, "{}").await?;
```
frondeus, line 168: I think we should insert more objects to really see whether this is inner join or outer join

e2e/src/on_demand_materializer.rs
luke-biel: Well, it's not `one_of options` combinator, it's `all_of options`, if possible of course

crates/dto/src/materialization.rs
```diff
@@ -302,6 +302,7 @@ impl Relation {
 #[serde(rename_all = "snake_case")]
 pub enum FieldDefinition {
     Simple {
+        // TODO: Consider adding optional base field - for fields from related schema without need of using subobject - currently workaround as computed field
```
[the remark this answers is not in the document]
Kononnable: I was planning to go through all TODOs(write them into issues if necessary) when phase II is close to being finished(all major features are implemented, docs are more or less there - scope is well known). There are a couple of TODOs left, some may get invalidated once other parts of phase II are finished.

Is it ok with you to keep it untill then?

e2e/src/on_demand_materializer.rs
Kononnable: Yes, like all other changes related to materialization. However, materialization documentation is blocked until its scope is known(until we decide what is a feature and what is not and how to group materialization feature documentation). This was partially solved by #608, #645 contains what is left(splitting features). After merging #608 at least the writing of part of the documentation can be done.
In general, we should document things sooner, I decided to add those tests before documentation so the whole partially implemented feature is not completely blocked and can be somewhat tested during implementation.

e2e/src/on_demand_materializer.rs
Kononnable: 2 objects are enough to tell if it's inner or any of the outer joins. It wouldn't be enough to be specific about which outer join strategy is used.
I don't see any more asserts necessary here, but If you think we need something more please add an example of a situation that this test misses.

crates/dto/src/materialization.rs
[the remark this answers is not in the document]
luke-biel: Yeah, if you are taking responsibility of doing this, it's ok with me.

e2e/src/on_demand_materializer.rs
```diff
@@ -0,0 +1,1169 @@
...
+        add_edges(relation_id, object_id_a, &[object_id_b]).await?;
+
+        sleep(Duration::from_secs(1)).await; // async insert
+
+        let view_data = materialize_view(view, &[schema_a, schema_b]).await?;
+        assert_eq!(view_data.rows.len(), 1);
```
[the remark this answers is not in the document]
Kononnable: and it's even a reason why this test was skipped :|

e2e/src/on_demand_materializer.rs
Kononnable: I have split the module into separate files. I decided not to go macro way for now - we plan to remove that code sooner and reading more verbose is easier for people unfamiliar with rust(not sure who will rewrite those tests into a new testing framework).

luke-biel (approved):

frondeus (approved):

Issue #615, opened 2021-07-02
general CI setup and related topics


esavier:
tested jenkins but with a pletora of problems we decided its too much to handle by us,
currently testing
- [x] goCI
- [x] laminar
- [x] jenkins
- [x] buildbot
as a result there wil lbe rfc describing problems and behavior for each tool

there are two expected results:
- [x] we have general purpose CI/CD coupled with github that will run basic acceptance and e2e tests checking if code changes are ready to merge
- [x] we have specialized CI/CD that allows us to run performance tests and benchmarks on known machine setup

currently todo from Playground perspective:
- [x] kafka setup
- [ ] rabbitmq setup
- [x] postgres setup
- [x] victoria metrics
- [ ] storage adjustments (ssd)
- [x] firewalls and security
- [x] connectivity and networking
- [x] k8s and nodes
- [x] logging, prometheus and graphana
- [x] access to the playground for users

currently todo from hephaestus perspective ( goCI setup )
- [x] goCI setup
- [x] certs and security
- [x] CI access
- [x] standard agents
- [x] elastic agent plugin
- [x] elastic agent config
- [x] basic build test

PR #640, opened 2021-07-13
docs: rfc: CI/CD setup R&D

esavier:
related #615


docs/rfc/NAF-CI_CD_setup_rnd.md
```diff
@@ -0,0 +1,45 @@
...
+## Tooling:
+
+in terms of availability and usability, following tools were tested:
+- Jenkins
+- Buildbot
+- Lamianr
+- GoCD
+
+#### Jenkins
+It is very versatile tool that allows for adjustments and modifications. Community is quite active and thre is a lot of plugins for different systems and usecases. Hoewver in our case there were problems with proper configuration of the service itself. Numerous issues with HTTPS reverse proxy (among others) was noticed and prevented us from, for example, providing or changing configuration and communicating with the agents. Despite that the setup can be adjusted to take that into account, for example by providing proper vpn connections, switching to plain http and locking the machine itself from outside world. Anoher problem is that java errors are not easy to understand nor to fix.
```
luke-biel, line 33:
```suggestion
It is very versatile tool that allows for adjustments and modifications. Community is quite active and there is a lot of plugins for different systems and usecases. Hoewver in our case there were problems with proper configuration of the service itself. Numerous issues with HTTPS reverse proxy (among others) was noticed and prevented us from, for example, providing or changing configuration and communicating with the agents. Despite that the setup can be adjusted to take that into account, for example by providing proper vpn connections, switching to plain http and locking the machine itself from outside world. Anoher problem is that java errors are not easy to understand nor to fix.
```

luke-biel (changes requested):
A typo, rest looks ok

docs/rfc/NAF-CI_CD_setup_rnd.md
```diff
@@ -0,0 +1,134 @@
...
+- Reverse Proxy - both load balancing and security element in network communication, usually http - related. In our case, its nginx. Its main job is to provide TLS wrapping on the http layer (resulting in https TLSv1.3). While load balancing is not currently a problem. it is a security concern to allow users to authenticate over plain text, and thus component like this have to be introduced to help alleviate this problem.
+
+## What we want to accomplish by introducing the CI/CD?
+a.k.a.  What's the situation, and what are we trying to solve?
+
+Currently, all our tests are being ran on the GitHub public cloud. We do have, as a OSS project, a lot of flexibility in this matter and GitHub provides for us almost infinite resources to help with our development. However, being the generic system, we are using VM's and environments that we do not control fully. We cannot guarantee that tests that we want to do will run on the same environment and thus tests are comparable with each other.
```
frondeus, line 21:
```suggestion
Currently, all our tests are being run on the GitHub public cloud. We do have, as an OSS project, a lot of flexibility in this matter, and GitHub provides us almost infinite resources to help with our development. However, being the generic system, we are using VM's and environments that we do not control fully. We cannot guarantee that tests that we want to do will run in the same environment and thus tests are not comparable with each other.
```

frondeus: Besides that, looks good to me

frondeus:
Please rename RFC filename to for example:
`0015_CI_CD_setup_RnD_01.md`

luke-biel (approved):

docs/rfc/NAF-CI_CD_setup_rnd.md
```diff
@@ -0,0 +1,134 @@
...
+
+Measuring itself is currently free-floating problem that does not have to be immediately solved. To be specific, it's the problem of amount of work and path we will take given setup that we like, but it does not have to, at this point of time, to dictate setup itself.
+
+### Solution:
+Due to us being partially lazy and partially having limited resources for our disposal, we decided that it would be the best idea to follow a hybrid solution for now. That means that we will keep GitHub actions for whatever we are doing right now, i.e. code quality and compliance testing, that can be run multiple times in parallel, and thus unlimited resources GitHub provides would be the best option for it. For performance testing, we will use the playground server stack which is configured, stable and persistent across tests. That also means that each test will be started in the environment as similar to each other as possible.
```
Kononnable, line 29:
```suggestion
We decided that it would be the best idea to follow a hybrid solution for now. That means that we will keep GitHub actions for whatever we are doing right now, i.e. code quality and compliance testing, that can be run multiple times in parallel, and thus unlimited resources GitHub provides would be the best option for it. For performance testing, we will use the playground server stack which is configured, stable and persistent across tests. That also means that each test will be started in the environment as similar to each other as possible.
```
This has nothing to do with laziness or limited resources(in a reasonable way). Github actions have some benefits that private hosted CI does not. E.g. CI status and run details are available publicly, any contributor can run tests on CI, even on their own fork - you don't need access to the internal environment.

docs/rfc/NAF-CI_CD_setup_rnd.md
```diff
@@ -0,0 +1,134 @@
...
+
+In addition, we will be able to perform server tuning and advise for best HW and SW configuration for CDL to run on.
```
Kononnable, line 31: What does `HW`, `SW` stands for?

Kononnable (commented):
I don't agree with all Architecture Considerations(e.g. need of tinkering with the kernel), but as it's research, not resulting in any direct decisions it's fine with me.

However, those two comments probably should be taken care of before merging.

docs/rfc/NAF-CI_CD_setup_rnd.md
frondeus: Hard Ware and Soft Ware I presume?

docs/rfc/NAF-CI_CD_setup_rnd.md
esavier: yeah i wrote that because i was lazy

Kononnable (approved):

docs/rfc/NAF-CI_CD_setup_rnd.md
```diff
@@ -0,0 +1,136 @@
+
+```
+Title           : CI/CD setup - R&D
+Author(s)       : Mateusz 'esavier' Matejuk
+Team            : CommonDataLayer
+Reviewer        : CommonDataLayer
+Created         : 2021-07-13
+Last updated    : 2021-07-19
```
Kononnable, line 8: I guess it's should be updated again

Kononnable (dismissed):

frondeus (approved):

frondeus (changes requested):
Please fix filename

esavier:
done, cheers mate :+1:

frondeus (approved):

PR #650, opened 2021-07-21
fix: materialization filtering

Kononnable:
Fix issues around materialization (filtering, fetching multiple objects)

As for the helper script - I can keep it locally if you want, however, I think it might be useful to some of you.


crates/schema-registry/src/rpc.rs
```diff
@@ -410,6 +410,11 @@ impl SchemaRegistry for SchemaRegistryImpl {
 
         let view = self.db.get_view(id).await?;
 
+        let filters = match view.filters.0 {
+            Some(f) => Some(f.try_into_rpc()?),
+            None => None,
+        };
+
```
luke-biel, line 417: Probably you can .map().transpose()? this

luke-biel (approved):

luke-biel (approved):

PR #652, opened 2021-07-22
chore: add group_imports = 'StdExternalCrate' to rustfmt

luke-biel:
This is a proposal, not to be merged before other functionalities are dangling (reapplying rustfmt is easier than rebasing branches). We already do that in half of our files.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+group_imports = "StdExternalCrate"
--- a/benchmarking/src/upload_to_kafka.rs
+++ b/benchmarking/src/upload_to_kafka.rs
@@ -1,3 +1,7 @@
+use std::io::Stdout;
+use std::sync::Arc;
+use std::time::Duration;
+
 use anyhow::Context;
 use clap::Clap;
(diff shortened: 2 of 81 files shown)


Kononnable:
https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=group_imports

Fine with me. One thing to note here - CDL crates will be grouped together with external dependencies.

>To enable unstable options, set `unstable_features = true` in `rustfmt.toml` or pass `--unstable-features` to rustfmt.

This option is unstable, so I think we should enable unstable features in `rustfmt.toml` instead of passing the command line parameter each time.

luke-biel:
From running this feature, it appears as if I don't have to pass unstable_features flag. Do we set it anywhere in CI?

Kononnable:
If it can run without passing the command line parameter I'm good with it. Maybe rustfmt docs are not up to date or unstable options are enabled on nightly builds by default or something similar.
I don't recall setting it anywhere.

luke-biel:
Close for now. Will reapply change later.

PR #653, opened 2021-07-22
fix: group by on unsorted elements in ER

Kononnable:
`group_by` groups only consecutive elements. Similiar error was discovered in OB(#650).

Some logic moved to make sure sorting the items is done in the same function as grouping(which requires sorted data).


crates/edge-registry/src/lib.rs
```diff
@@ -299,20 +299,24 @@ impl EdgeRegistryImpl {
...
-        &self,
-        object_id: Uuid,
-    ) -> anyhow::Result<impl Iterator<Item = (Uuid, Uuid)>> {
+    async fn get_edges_impl(&self, object_id: Uuid) -> anyhow::Result<Vec<Edge>> {
         counter!("cdl.edge-registry.get-edges", 1);
         let conn = self.connect().await?;
         Ok(conn
             .query(
-                "SELECT relation_id, child_object_id FROM edges WHERE parent_object_id = $1",
+                "SELECT relation_id, child_object_id FROM edges WHERE parent_object_id = $1 ORDER BY relation_id",
```
frondeus, line 307: I'm not sure whether we could use simply GROUP BY in SQL.

However, it might have some performance implications because under specific load there might be more client, backend CPU time than SQL Postgres CPU time, and therefore it is more performant to do grouping in Rust.

Kononnable: I'm fine with using `ARRAY_AGG` function to group `child_object_id` column. However, it requires some Postgres(not general sql) knowledge - if the team is fine with that I can change it.

As for performance - the general rule is to work on data as close as possible, especially if we're processing the whole dataset(not row by row - cursor-based approach). Grouping on SQL can be done without the need for sorting beforehand(same on rust but SQL engine will choose the fastest path based on the number of rows and other parameters). Anyway, I wouldn't treat performance as a priority here for now - code readability is more important, and we can always change the internal workings if it becomes a bottleneck.

luke-biel (approved):

PR #654, opened 2021-07-22
fix: change fetch-many to fetch in schema_registry/db.rs

luke-biel:
To test:
* add 2 schemas
* add relation schema_a -> schema_b
* add view with relation
* query GetAllViewsByRelation over gRPC (schema-registry)


frondeus:
Maybe @pgr-automation  could write an automatic test that checks it?

Kononnable (approved):

frondeus (approved):

PR #661, opened 2021-07-23
fix: drop mutexes around NotificationPublisher

luke-biel:
It appears that these were added on accident. Also, and_message_body doesn't have to take self by value. We would need to clone Self interia anyway.


frondeus (approved):

crates/utils/src/notification/mod.rs
```diff
@@ -51,16 +51,16 @@ where
     T: IntoSerialize<S> + Send + Sync + 'static,
     S: Serialize + Send + Sync + 'static,
 {
-    pub fn with_message_body<U>(self, msg: &U) -> Box<dyn NotificationService>
+    pub fn and_message_body<U>(&self, msg: &U) -> Box<dyn NotificationService>
```
Kononnable, line 54: Why `and` instead of `with`? I think `with` suggests some kind of a payload, `and` suggests that something will happen after(`and_then`).

Maybe we should rename it to something like `to_notification`? Returned `NotificationService` isn't something general - it works only with this single message.

Just a thoughts, doesn't block merging in any way.

luke-biel: `with` is considered consuming or mutating in rust (`self` or `&mut self`), `to` and `into` are converting prefixes. And was only thing I could think of.

PR #663, opened 2021-07-26
feat: Error reporting for materialization

### LINKED ISSUE #543: Object builder and Materializer error handling
Both OB and MG have to send notifications to user once a row fails to build or get inserted.

**Notes:**
May require custom solution rather than notifications that we currently use in cdl.

**AC:**
* OB writes to error channel when it fails to create materialized view row
* MG writes to error channel when it fails to insert a row into target database

frondeus:
fixes #543


.cdl/development/schema-registry.toml
```diff
@@ -1,4 +1,5 @@
 input_port = 50101
+import_file = "bare/setup/schema-registry/initial-schema.kafka.json"
```
frondeus, line 2: We need a nice way to patch those entries on horust level, maybe using `patch`? Or creating a more sophisticated `overrides` mechanism?

frondeus: For now either
```suggestion
#import_file = "bare/setup/schema-registry/initial-schema.kafka.json"
```
or
```suggestion
```

crates/partial-update-engine/src/main.rs
```diff
@@ -48,10 +48,12 @@ struct ServicesSettings {
 }
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Hash)]
+#[allow(clippy::enum_variant_names)]
 #[serde(untagged)]
 enum PartialNotification {
```
frondeus, line 53: We could extract it into DTO and reuse in API, perhaps

.cdl/development/schema-registry.toml
luke-biel: That's y I would prefer for everyone to create their own set of configs. I dunno, maybe Friday meeting will shed some light onto this. I'ma note that this is an use case.

Kononnable (commented):
Looks like I forgot to press submit review button :|

.cdl/development/schema-registry.toml
esavier: its problematic for horust deployment to have some additional steps before launching, config management left on user side is a problem since  you have technically have to make matrix consisting of (deployment * version * variant).
I was under the impression we will be able to make a directory with variants in the cdl repo so that we can patch it up against deployment automatically during the start.
I dunno, we will have to think something out.

esavier (commented):
lets wait for comment resolutions

crates/object-builder/src/lib.rs
```diff
@@ -67,63 +71,7 @@ pub struct ObjectIdPair {
     pub object_id: Uuid,
```
smores56, line 71: If this type has its own module, its definition should be moved there.

crates/object-builder/src/lib.rs
```diff
@@ -282,20 +241,34 @@ impl ObjectBuilderImpl {
...
+                let report_sender = report_sender.clone();
+                async move {
+                    let (rows, errors): (Vec<RowDefinition>, Vec<String>) = rows
+                        .into_iter()
+                        .map(|r| r.map_err(|e| format!("Could not build a row: {:?}", e)))
+                        .partition_result();
+
+                    for error in errors.into_iter() {
+                        tracing::error!("Could not materialize row: {}", error);
+                        report_sender.clone().notify(error.as_str()).await?;
```
smores56, line 262: This seems fragile. What happens if we fail halfway through? If this is the desired behavior, maybe we should document our approach (both here and in our docs?).

crates/object-builder/src/lib.rs
```diff
@@ -43,7 +46,8 @@ type DynStream<T, E = anyhow::Error> =
 type ObjectStream = DynStream<(Uuid, Value)>;
 type SchemaObjectStream = DynStream<(ObjectIdPair, Value)>;
 type RowStream = DynStream<RowDefinition>;
-type MaterializedChunksStream = DynStream<MaterializedView>;
+type MaterializedChunksStream =
+    Pin<Box<dyn Stream<Item = anyhow::Result<MaterializedView>> + Send + 'static>>; // Why this cannot be Sync?
```
smores56, line 50: This comment should be removed before merging.

crates/materializer-general/src/lib.rs
```diff
@@ -117,6 +109,36 @@ impl GeneralMaterializer for MaterializerImpl {
...
+}
+
+impl MaterializerImpl {
+    fn validate_options_inner(&self, options: &str) -> anyhow::Result<()> {
+        let options: serde_json::Value = serde_json::from_str(options)?;
+        self.materializer.validate_options(options)?;
+        Ok(())
+    }
+
+    #[tracing::instrument(skip(self))]
```
smores56, line 137: Why do we skip `self` here and not in the above function? Because this one is `async` and the other one is not?

crates/materializer-general/src/lib.rs
```diff
@@ -117,6 +109,36 @@ impl GeneralMaterializer for MaterializerImpl {
...
+    async fn upsert_view_inner(
+        &self,
+        materialized_view: MaterializedView,
+    ) -> Result<(), tonic::Status> {
         tracing::debug!(?materialized_view, "materialized view");
 
         let error_handler = |err| {
```
smores56, line 144: If all of these fallible components use the same error handler, they should maybe be wrapped in a better-named helper function or a `try` block. https://yaah.dev/try-blocks

smores56 (changes requested):

crates/materializer-general/src/lib.rs
frondeus: By `above function` do you mean `validate_options_inner`?

crates/object-builder/src/lib.rs
frondeus: Okay, Ill rename module to `object_id_pair_serde`

crates/materializer-general/src/lib.rs
smores56: Yes

frondeus: Okay, then the answer is - not really.
First of all, you add `#[tracing::instrument]` to attach span to the function.
It does not matter whether it is a method (receiving `self` argument) or async (returns Future).

This span can be sent to, for example, jaeger so you can see when the function was called, how long it took, and what events were created (for example, `tracing::error!()` is one event, `tracing::info!()` is another.

This attribute macro also attaches context - each argument passed to the function is serialized into the span. Serialization usually is achieved by using the `Debug` trait.

Now, `skip(self)` means, `do not serialize self when you create a new span, ignore it` just like you can `serde(ignore)` field of the structure.

So the question should be - why are we not attaching span into `validate_options_inner`, and we do for `upsert_view_inner`. 

And the answer is - because we didn't have to. Validate options do not produce a single log (event); it is unused for now (it's a stub of future implementation), and we do not care right now how long it takes (recording span) because it is a stub after all.

`upsert_view_inner` on the other hand, takes some time (it writes to DB) which we would like to measure and see in jaeger, it generates at least one event (see line 149) which we want to link to this function, and therefore I attached `tracing::instrument` to it.

smores56: Okay, thanks for the perfect explanation!

smores56 (approved):

smores56 (approved):

Issue #636, opened 2021-07-12
Specify definition of BreakingChange in cdl


luke-biel:
Some of our apis are still immature and it would be difficult if we'd make them frozen until 2.0 release. We should review what explicitly relates to breaking change and what can be changed on demand.

Issue came up on SR proto file change. In theory we can say, that intended way of accessing SR is via cdl-cli and live with it for a time being. Same probably goes for other components, QS and CS don't need to have apis frozen. QR has to, but maybe only to some extend.

DoD:
* A document is present, in which we have explicit definition of apis that are final for the time being, and classification criteria are specified.

Kononnable:
Related: https://epiphany-platform.github.io/CommonDataLayer/versioning.html#breaking-changes

Kononnable:
Examples of what to clarify:
- [ ] Hiding feature under configuration flag(disabled by default) - a client can avoid breaking behavior by updating deployment configuration
- [ ]  Changes to SR
- - [ ] grpc contract change(changing field name)
- - [ ] changing schema/view definition format

Issue #567, opened 2021-06-18
chore: update developer deployment - alternative setups


frondeus:
Acceptance criteria:

* [ ] There is separate configuration (in `.cdl` directory) for GRPC
* [ ] There is separate configuration for RabbitMQ
* [ ] There is documentation how to switch between different setups
* [ ] (optional) Horust deployment is up-to-date and compatible with the configuration.

esavier:
643 is not fixing the issue

luke-biel:
Conclusion of discussion:
* we will be keeping .cdl subfolder with postgres + kafka/grpc/rabbitmq configs
* we will add workflow that will check whether .cdl is up to date (generator requires non-interactive run method for that)

PR #666, opened 2021-07-27
docs: rfc: Configuration Service

### LINKED ISSUE #224: Configuration Service - rfc
We need to create a Configuration Service that dynamically updates CDL configs.
This task is about creating an rfc covering the topic.

I will use `CR` shorthand from now on, `Configuration Registry`. configuration layer can use one name, registry (SR, ER, CR).

Some notes:
* if config is dynamic, which configs should be refreshed by apps? some cases, like change of communication method, require substantial work thats equivalent to restarting a service, so some configs may remain "static"
* we have to figure out a way to easily integrate dynamic configs into our apps; android ViewModel observe/angular observables may be good starting point for the design
* we have to figure out how `CR` should propagate config updates
* can there be a `CR-less` environment? should we allow cdl to startup without CR up
* should we keep Config.toml, so that CR is a layer of configuration on top of static files?
* do we need CR? to what extent? what problems are we solving by providing a dynamic config?

AC:
* An rfc about architecture is created, discussed and accepted/rejected. It covers where `CR` propagates configs, where it lies within CDL domain and how CDL reacts to propagated configs. Some concrete implementations may be proposed, but it should mainly discuss higher level design.
* An rfc about CR implementation is created, discussed and accepted/rejected. There are many ways to implement a solution, we should review them all. There are some PoCs brewing right now, let's take them into account.

DoD:
We are presented with set of document and decisions that allow us to create a CR implementation.

esavier:
fixes #224


docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Introduction
+It is required for CDL to handle dynamic configuration and provide configuration point for the deployment.
+
+Independent of the implementation chosen, below are the major restrictions the configuration service has to comply with.
+
+#### - requirements:
+- providing central configuration repository
+- providing configuration for specific services
+- allowing services to be configured by group or instance
```
Kononnable, line 47: Not sure I understand this one correctly.
I don't think we need the ability to configure 1 of 2 instances of DR to work in some way. It will always be all of DR instances or none of them(configuration by service group, not instance).

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+#### - optional requirements: 
+- providing dynamic configuration
+- providing proper authentication and authorization of configurations
+- allowing for service reconfiguration on configuration change
+- allowing for high-level service management
+- service discovery and autoprovisioning
+
+#### - additional assumptions:
+- solution has to be neutral against deployment
```
Kononnable, line 59: This is something different from the last point from requirements or just a repetition to make sure everyone is aware of this?

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+- solution has to provide api that is permissive to 3rd party services which may be written in different languages.
+
+#### Additional notes:
+As this is a general RFC, most functionality that is taken into consideration does not have to implemented right away, however it is crucial to both describe it and take it into consideration during feature planning.
+
+#### - dynamic configuration
+While notifications are optional, its availability can influence deployment's behavior. For example, in case of lack of dynamic configuration, the whole deployment will have to be reset and started anew. The same have to happen in case of nonexisting notifications in some cases of dns-less deployments, for example when service will crash and restart with different port, or in case of hot-adding services during the runtime.
+
+High level service management is something is not utilized yet, but may be necessary to provide it in the future. The Simplest example of which is to restart the service on demand by operator (administrator, for example). Such behavior is not required right now, however the choice of solution should be influenced by its possibility to support that feature in the future.
```
Kononnable, line 68: I think this feature may be nice to have, but it's not really possible to implement without limiting users' deployment strategy in any way.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+-- CFGS's general private key
+-- operator's private key
+-- list of secrets bound to particular functionalities or access groups
+- Client of configuration service has to have:
+-- operator's private key
+-- CFGS's public key
+-- client's private key
+-- secret used to identify the permissions of the client
+
+Additionally, the payload have to be protected by hash authentication code (i.e. AES-GMAC) in the future. This is the mechanism that protects both the message and its hash with control sum, which is first appended then encrypted together with the message.
```
Kononnable, line 82: One of many options presented as the only solution. Not sure why those are the criteria you have chosen, they're probably the reason why some of the solutions below are marked as 'no security/no authentication' even though they provide methods for both(even HTTP protocol has authentication support https://datatracker.ietf.org/doc/html/rfc1945#section-11).

I'm not saying that just http basic authentication is enough, but the World Wide Web is fine with HTTPS and sending logins and passwords in plain text for authentication purposes.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+Note: if Server sends notification with those tags, only Client1 should react, at least until filters are not implemented. With simple tags, only matching tags are going to react, however tag configuration needs to be cautious not to mix allowed tags. Further implementation can change the format to allow for secrets, so only authorized services can use specific tags.
+
+A second option that is viable in that case is topic separation, although the application will have to listen to multiple topics, which may not be optimal depending on the implementation and library behavior. However, little research was done to prove that, so it is left for implementor to decide.
+
+Configuration Service will be treated as an internal API and changes to it will not be considered breaking change. 
+
+It is acceptable to define API during early development stage and change or expand it later.
+
+Integration with GraphQL or management panel has to be considered in later stages of the implementation to provide API for human end users.
```
Kononnable, line 115: Not sure if we should discuss it now or later on - our GraphQL API is user-centered. It has no admin configuration so far, so it may be wise to split admin configuration(e.g. topic names) from user ones(schemas, views).

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+Some base work has already been done in separate [repo][magnetite], it can also be treated as a result of the research.
+
+
+#####  References:
+ CDL :
+[CDL project](https://github.com/epiphany-platform/CommonDataLayer)
+[CDL - RFC discussions](https://github.com/epiphany-platform/CommonDataLayer/discussions/categories/rfc)
+[CDL - RFC candidates](https://github.com/epiphany-platform/CommonDataLayer/tree/develop/docs/rfc)
+[CDL - RFC releases](https://github.com/epiphany-platform/CommonDataLayer/tree/main/docs/rfc)
+[1](¯\_(ツ)_/¯)
```
Kononnable, line 269: (¯\_(ツ)_/¯)

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+- in case multiple configuration services are to be used, a replication and consent algorithms have to be implemented
+
+Summary:
+- relative work: 5/10
+- maintenance work (periodic): 2/10
+- flexibility of the solution: 5/10
+- available features: 5/10
+- work needed to expand in the future: 5/10
+
+### P2P network
```
Kononnable, line 214: How does changing the value work in the proposed p2p solution: can any client change the value of data or only the owner(CFGS)? 
Can we scale CFGS service or it can only have one instance(one owner of data) but the load is spread over the network(only passive replicas are mentioned here)?

Kononnable (commented):
Didn't go with the review of each proposed solution yet. Some of my problems with them are related to other mentioned issues(e.g. security section). Just two questions about p2p solution for now.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+#### - requirements:
+- providing central configuration repository
+- providing configuration for specific services
+- allowing services to be configured by group or instance
+- providing at least a minimum degree of the security
+- configuration provisioning and configuration service should be abstract from the way CDL is deployed and should not limit deployment in any way.
+
+#### - optional requirements: 
+- providing dynamic configuration
+- providing proper authentication and authorization of configurations
```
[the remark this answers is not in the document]
esavier: 2FA is authorization confirmation method, not authorization itself. In case of 2FA login for linux, pam is responsible for your authorization based on 2FA you provide.
in that case authorization is about admitting connection to do specific work, i.e. changing the configuration.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+It is required for CDL to handle dynamic configuration and provide configuration point for the deployment.
+
+Independent of the implementation chosen, below are the major restrictions the configuration service has to comply with.
+
+#### - requirements:
+- providing central configuration repository
+- providing configuration for specific services
+- allowing services to be configured by group or instance
+- providing at least a minimum degree of the security
+- configuration provisioning and configuration service should be abstract from the way CDL is deployed and should not limit deployment in any way.
```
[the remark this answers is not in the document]
esavier: requirements were checked with PO,
if you will do it against cdl that is supposed to work only on k8s then it won't work anywhere else.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+#### - optional requirements: 
+- providing dynamic configuration
```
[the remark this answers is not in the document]
esavier: first version does not have to provide dynamic configuration, worst-case scenario it will be taken care of in the expansion of the initial CFGS, 
There are a lot of problems related to dynamic configuration, some of which are explored in depth in the document. Providing all cases and solutions for those, where dynamic config is problematic, would expand the document to twice the size.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: This is related to access groups.
There is for example deployment case where we have two different possesses and command services for those require different configs (so two groups of similar configs)

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: yeah this shouldn't be here

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: it's how the p2p approach is doing things, that's why it's recommended.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: http authentication model is session authentication, which is not secure.
I believe that any kid with a keyboard can break this.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: This i am not sure how to tackle, so probably you are right, meeting is needed, but in worst-case scenario we will put second GraphQL service for that.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: because there are changes in the rfc naming etc, its waiting

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: 1. no, clients, can report their change (which is not described in this document). In first release, CFGS is a source of Truth. 

2. it can make replicas, both passive and active, so spreading the load is possible, although some code will be required. Replicas do not replicate automagically, but there is a basic consensus mechanism available. Load can be spread across instances if needed.

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
Kononnable: Even if the first version is not required to provide dynamic configuration, but it will be required later on I don't think we should choose technology that doesn't allow it. Otherwise, we would have to rewrite the configuration service completely then.

That's why I don't think this requirement is optional for choosing technology behind the CFGS even if it's optional for the first version of the service.

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
Kononnable: 2FA is 'part of' Multi-factor authentication, it's about authentication, not authorization.

It can be treated as a confirmation method, but it doesn't change the fact that it's part of the authentication process, so if we want to take care of authentication we also have to decide if we support 2FA, what authentication methods do we support, etc.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: Access groups as described some time ago allows for multiple tenants to share the environment and resources. It doesn't start a separate set of services for each tenant, it uses application logic to serve multiple tenants in a single service instance.

What you're referring to sounds like two separate CDL deployments sharing single configuration service.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: I think we're supposed to choose the best option for CDL. I wouldn't say everything about how p2p works will work well for CDL.

And I think it doesn't change the fact that implementing high-level service management is impossible without limiting user deployment strategy - which is marked as a hard requirement earlier in the document.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: WWW currently uses multiple authentication models and most of them are not classified as insecure. This paragraph suggests that only using described method provides enough security for CDL which I don't think it's true and using this specific authentication model may not be a requirement when it comes to CDL security.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: Ok, so if it is either a multi-leader or a leaderless approach how is the problem of split-brain handled?
We cannot allow data corruption or data inconsistency here, so the resolution of the split-brain problem may be critical here.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: I don't think any changes will affect your branch unless you merge them in(and fix the link then).

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
esavier: As an item, it is optional, but i know and i do agree that taking a path that prevents this to be implemented somewhere will be problematic. 
I will confirm with @pljadru if its okay.
I can also rephrase it if its okay (i.e. rename the section, but i do not know how thought) or make a comment.

smores56 (changes requested):

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+Summary:
+- relative work: 3/10
+- maintenance work (periodic): 5/10
+- flexibility of the solution: 3/10
+- available features: 3/10
+- work needed to expand in the future: 6/10
+
+### REST server
+It is a mutation of the previous method. Server is a custom service that serves configuration in request-response model. Additionally, the server can support both mutability and notifications. Services interested in the configuration will have to ask for server directly for required configuration.
```
smores56, line 182: It isn't the method directly above, so I would say "It is a mutation of the above (or aforementioned) method" instead.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+Implementation of notifications will be problematic since there are two ways of achieving that:
+- Passive way: that there will have to be some kind of method to keep connection to the services. If we consider big deployments, for example running hundreds of services that have to be configured, then the configuration service will have to keep connection to all of them. Either streaming (hanging connections) or reversing client/server roles can be used to achieve that.
+-- First option is to use a mechanism that allows client or server to notify the other side that more data is on the way, however the problem is that it cannot be held indefinitely. Additional code will have to be put to provide proper medium and its behavior, and it will have to be tested separately.
+-- Second option means that Server will act as a client and Client will act as a server. Configuration service will send the notification to servers that will look like a simple request with all necessary changes. This method, however, requires us to implement some additional code to register the service first, and a lot of additional work with securing the channels.
```
smores56, line 187: This is not being rendered properly as a sub list by GitHub. You should use `  - ...` instead. (2 spaces in front)

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+Summary:
+- relative work: 5/10
+- maintenance work (periodic): 2/10
+- flexibility of the solution: 5/10
+- available features: 5/10
+- work needed to expand in the future: 5/10
+
+### P2P network
+In this section, libp2p was taken into consideration as way-to-go library. P2P network can alleviate most of the programmer-side issues that are explained in previous section, while providing transparent that is scaling with the deployment itself. Libp2p works by providing the implementor with a set of default "network behaviors" that dictates how the node should behave in various situations, examples of which being particularly important to the subject at hand are "gossipsub" and "request-response". Explanation of those behaviors are below:
```
smores56, line 215: Transparent what? Did you mean to say transparency here?

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+- gossipsub - more advanced publisher subscriber model (pub-sub being a best suited model for any p2p and/or meshing networks). One limitation of this model is the requirement to have at least 3 services running. General characteristics of the behavior is :
+-- Services are building the network automatically and keeping only few connections to nearby peers, building a "mesh".
```
smores56, line 217: Again, please use GitHub-compatible sub lists.

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+- low maintenance, done right at the start, it 
+- p2p is transparent for the user
+- network autoconfiguration and autodetection of services out of the box
+
+neutral:
+- there is more work to be done to achieve reasonable degree of authorization
+- libp2p implementations are available for major languages, but not all of them
+
+negatives:
+- lot of work initially
```
smores56, line 242: I'd say that even more negative than the initial front-loading of work is the issue that we don't know how long this will take. Should that be mentioned here?

docs/rfc/CDLF-0000F-00-rfc-01.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Addendum
+Both p2p request-response and general rest's major problems is related to the peak load. Those issues can be alleviated in relatively similar way, both client-side and server side. Examples of which are, implementation of random sleep periods while acting on the notifications, or having multiple configuration services available. Those options will not be explored in this document because those are tightly coupled with implementation itself.
+
+While compatibility with other libp2p libraries written in different languages is stated in documentation and implied by following the same spec, it was not tested in the field by the author of this RFC.
+
+## Chosen solution
+I, personally, would opt out for libp2p solution, as the investment seems to be worth it. The amount of work required is quite high, however it can be treated as an investment, since after initial feature release is ready, almost no additional work have to be done. P2P part only have to be wrapped in common library. It is providing automatic, invisible layer to the users, and it does not have to be expanded further. All feature-related code is then common between the solution with the rest server.
```
smores56, line 258: To opt out is to choose not to do something. You probably meant opt in (to choose an option with what it entails).

smores56 (changes requested):

smores56:
I like this document, and think that most of my discussion points with it will be relevant when we meet to discuss it, more so than now, where I have mainly only seen minor grammar issues. Good job!

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
esavier: Authorization is idea that is kind of larger than authorization itself, if we are talking about users, there is no proper authorization without authentication, and there is no reason for 2FA for authentication itself. 
Here we are talking about message authentication inside the deployment, not really about outside, user-facing security since we do not have this defined yet, additionally user-facing security will depend on the implementation which first have to be chosen. From my understanding, if we assume graphql, then there is not really much we can do, if a different approach is taken, then there are few options, however this document does not take any assumptions on that. 

For libp2p solution, internal security is given out of the box, with exclusion to authorization, for other solutions it has to be planned and implemented when the time for it comes.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: what is tackled in the paragraph is security on top of the transport layer inside cdl, as configuration changes can not be accepted from malicious services running in the same stack. User security is not described but its also out of the scope. It will be designed when there is time for that. Currently, there is only a need to have a place for it or. 
libp2p just delivers most of it out of the box, without testing nor explicit implementation from our side.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: you gave nothing to support your argument, this comment is so vague i do not how or what should address  here

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: thanks dad, for telling me how to merge my things

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: the same way as in any other solution,
if you interested in details, go and read either documentation or spec of libp2p, and you will know how to handle this

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: in high level i think it is, you are still serving static data over an endpoint with req-response model. How and in what server is written is completely secondary, may it be bash, nginx, zig or rust. Only difference in here is that some kind of mutability is preserved which in case of nginx can not

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: yeah, i tried to do this as fast as possible, i read this through twice but i missed few things,
i wanted to write "transparent communication channel"

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: we have PoC working separately, i think we gauged pretty neatly how much work we have to put in this,
1. [done] make req-resp work
2. [done] test req-resp
3. test req-resp + pub-sub
4. wrap it in the coherent module (*)
5. implement handlers for services (*)
6. implement configuration exchange proto (*)
7. implement synchronization (**)
voila

* can be done partially, i.e. mvp at this time and expanded later
* can be done when scaling is requested

esavier: also by a lot i do not mean "a lot", rather "a lot more, but its not so terrible"

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: > 
> 
> I like this document, and think that most of my discussion points with it will be relevant when we meet to discuss it, more so than now, where I have mainly only seen minor grammar issues. Good job!

Thank you very much, I worked really hard on this  ❤️ 😁

esavier:
> I like this document, and think that most of my discussion points with it will be relevant when we meet to discuss it, more so than now, where I have mainly only seen minor grammar issues. Good job!

Thank you a lot, i put a lot of effort in this ❤️

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
Kononnable: So what exactly is the optional requirement here? 
Safety of internal messages (e.g. configuration changes propagation) not the safety of user requests to change config value?

If so this should be clarified in the document.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: Again, it's not about specific proposed technology(libp2p), it's about what we require from any solution. If it's a description of libp2p it should go into its section.

For other technologies, different 'verification protocols' may be used if they provide enough security. We're not limited to using private keys.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: Different solutions use different methods, there is no silver bullet here. Some solutions even don't have this problem at all - they always have a single source of truth(but do not qualify for high-availability standards).
It's important as sometimes the split-brain problem is not handled at all on the protocol level, or is handled in a way that doesn't provide guarantees needed for configuration service.

I think this document is supposed to help to decide which way we want to go without everyone getting into specs of each proposed technology.

docs/rfc/CDLF-0000F-00-rfc-01.md
Kononnable: I'm just saying that if `the choice of solution should be influenced by its possibility to support that feature` then we should not say it's recommended to do this way because one of the solutions is doing it this way. We're here to pick a solution, not to argue why we choose a specific one(the decision was not made yet).

Anyway, we got off the topic here. Original problem was that operations like restarting services on-demand are something that cannot be done without limiting users' deployment strategy - which is specified as a requirement.

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
esavier: okay, i can stand behind that

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: technically AG doe not have proper rfc yet, we only had a sketch of the functionality. 
Technically we may have different (abstract) groups with different functionalities and access levels, its not clear yet. That means we will have to change properties per group not per type of service.
i.e. type of service does not match 1:1 with group, or for example we may have two postgres instances with different functionalities enabled.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: i kinda disagree, we can restart the services both on baremetal and on container management platform, which are two, most basic ways to deploy the CDL. Other options, like istio or linkerd are variants of the second approach. 
In case of both deployments, the easiest way would be to gracefully exit the application, and reload with new config, and this is possible both in case of baremetal with proper functionalities (like proper systemd service file) and with container mangement.

Kononnable:
Discussions about requirements were clarified(solved or we know what exactly we have problems with), so it's probably a good time to add more solutions to choose from.


## File configuration instead of adding another service:
Actually, this idea was proposed a long time ago. Its core is that we put all of our configuration data into files. Single service can read from multiple files or all services reads the same big file(it's up to us on if/how we want to split configs, probably some splitting is a good idea to keep deployment setting separate from settings related to user application). Services watch for file changes(inotify) and act on them when files get changed(detect changes and apply them). We(as CDL) don't really have to care how files are replicated over the network however, we should provide some default options. In custom deployments, it's up to the ops team to decide which way of file synchronization is good for them(based on the requirements for that deployment - e.g. time for all services to read new config).
For k8s deployments, we can simply use config maps/secrets(if we assume CDL admin will have direct access to k8s) or use sidecars to fetch/synchronize files in a different way. Those sidecars can be custom-made for deployments which gives a lot of customization and flexibility for end-users(users can define the ultimate source of truth). We can add a new service on top of that configuration propagation if we need more programmatic API for configuration service or we want to validate configuration before it is applied.

One rough thing about this approach is configuration validation. With a file-based approach, configuration can't be validated at the moment of the change(we can't deny the change if the configuration is not valid). Services would have to know how to act when the configuration is corrupt - report errors and continue working on the older version. This approach would work well on already existing services(including service restarts) but is hard to achieve with instances created once invalid configuration is set(e.g. automatic scaling). 
To migrate the risks we can ask CDL admins to validate the configs before pushing them(config file validation is already planned as part of #567) and watch for reported errors after the change.
As for k8s, we can write a simple operator which job would be to validate the configuration settings and propagate them(update config maps) or utilize a separate new service mentioned earlier. This will not disrupt CDL's ability to scale if someone provides the wrong config file(the whole CDL would be running under old configs).

Pros:
- easy to implement/setup
- simplicity
- dynamic configuration
- secure(attacker needs direct access to k8s or depends on user-specific synchronization method)
- no split-brain problems - unless the user use a synchronization method that is vulnerable to this problem(not likely)
- customization and flexibility - users can completely change the behavior of 'config service' if they need to(they can easily add custom logic on top of that, use different source of truth), with other mentioned solutions that would require changes to CDL code
- no real resource runtime costs(default synchronization method uses k8s(etcd)/docker volumes/filesystem to propagate changes - no runtime costs of listening for change events)

Cons:
- configuration change is async(same as with others mentioned methods so far) - user has to manually check logs to see if configuration change was successful
- lack of configuration validation check on saving the new configuration


## Central service:
Each CDL service connects to a single(or replicated) configuration service and keeps the connection open(e.g. WebSocket). Configuration service provides API to end-user - it's up to us to decide if that should be REST/GRPC API or e.g. changing values stored in pg should be enough to trigger the change. Configuration service on change request validates new data and pass configuration change event to interested services. 
In theory configuration service may be seen as a single point of failure, however in this case it shouldn't be a big problem:
- K8s or other deployment methods can easily decide if service is unresponsible and restart it if necessary. Downtimes shorter than a minute are not problematic for configuration service as it's not serving the users directly.
- If any CDL service loses the connection to configuration service(e.g. network problems) it can still work and try to reconnect in the background. We may want to introduce a configuration parameter to stop the 'unresponding' service from working if it cannot connect with the configuration service for a specific amount of time(to make sure we're not running any service with old config).
- We can add replication to configuration service(raft or similar protocol) so even if one instance is unresponsible clients will be able to connect to another one thanks to load balancers

Pros:
- easy to implement/setup
- dynamic configuration
- automatic configuration validation
- secure(we are free to decide on the usage of any authentication/authorization protocols)
- no split-brain problems - configuration service(or its database) is a single source of truth, in case of replication we can use any protocol which handles that problem well(e.g. raft)
- configuration change can be sync(in case of REST/GRPC we can delay sending the response until all connected services acknowledge the change) - nice error reporting for scripting purposes(no need to check logs)
- small runtime resource costs - WebSocket and similar protocols have low resource costs, a single configuration service instance can easily handle thousands of connections(and it's unlikely that CDL will be scaled in thousands)

Cons:
- theoretical single point of failure(described earlier)

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: all kinds of security protocols will boil down to either symmetric encryption or asymmetric one. I think its a little to much to describe diffie-helman key exchange mechanism or different symmetric keys here. If its still the TLS you are drinking to, its a combination of both symmetric and asymmetric encryption wrapped inside a specific mechanism, problem with that is that we will have to provide proper certificates and certificate authentication mechanism to consider together with TLS wrapper itself. Technically TLS will work with CFGS and it covers the requirements, but its very problematic for local network, and that's why i did not propose it as a solution.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: libp2p protocol provides message identification mechanism, i.e. you can define which message happened first, additionally it provides a way to keep the knowledge about the last n messages on protocol level. Consolidation have to be implemented, as in any other approach, but we have consensus and knowledge about all the messages in proper order. It behaves literally the same way as a message queue system. 
On first stage consolidation does not have to be implemented, in case we want it to scale at some point, we will have to do that, also with libp2p (unless we will go with kademlia model which is not described but also a possibility).
Desynchronization of states is therefore not handled on protocol level. The protocol only provides facilities to implement it since consistent state is easy to achieve using limited network broadcasts. 
Thats why i deemed it not a problem for now, and that is also why i advised to read a specification of the protocols, it is really nice lecture to have.

esavier:
i am not sure but its amost the same i wrote on solutions 1 and 2, although with greater emphasis on it ?

Kononnable:
Yes, using files for configuration is similar to the first two approaches. However, because of differences(and some generalizations with different 'defaults'), most of the negatives are no longer valid or even converted into positives.

Central service is similar to the rest server approach, but again the more detailed design of some aspects allows to migrate most of the negatives.

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: done?

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+#### - Requirements
+- providing central configuration repository
+- providing configuration for specific services
+- allowing services to be configured by group or instance
+- providing at least a minimum degree of the security (protocol sercurity, i.e. no config changes from unauthorized sources, message verification etc)
+- configuration provisioning and configuration service should be abstract from the way CDL is deployed and should not limit deployment in any way.
+
+#### - Optional Requirements:
+- providing dynamic configuration
+- providing proper authentication and authorization of configurations
```
luke-biel, line 52: Do you mean access groups by this?

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+
+#### - Requirements
+- providing central configuration repository
+- providing configuration for specific services
+- allowing services to be configured by group or instance
+- providing at least a minimum degree of the security (protocol sercurity, i.e. no config changes from unauthorized sources, message verification etc)
+- configuration provisioning and configuration service should be abstract from the way CDL is deployed and should not limit deployment in any way.
+
+#### - Optional Requirements:
+- providing dynamic configuration
```
luke-biel, line 51: Literaly first sentence of this paragraph mentions this. It is confusing

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+#### - Additional Assumptions
+- solution has to be neutral against deployment
+- solution has to provide api that is permissive to 3rd party services which may be written in different languages.
+- user facing security is separate issue which is not addressed here, however solution has to allow for it.
+
+#### Additional notes
+As this is a general RFC, most functionality that is taken into consideration does not have to implemented right away, however it is crucial to both describe it and take it into consideration during feature planning.
+
+#### - Dynamic configuration
+While notifications are optional, its availability can influence deployment's behavior. For example, in case of lack of dynamic configuration, the whole deployment will have to be reset and started anew. The same have to happen in case of nonexisting notifications in some cases of dns-less deployments, for example when service will crash and restart with different port, or in case of hot-adding services during the runtime.
```
luke-biel, line 66: What have notifications that are sent by CS and ER to do with this+

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+
+High level service management is something is not utilized yet, but may be necessary to provide it in the future. The Simplest example of which is to restart the service on demand by operator (administrator, for example). Such behavior is not required right now, however the choice of solution should be influenced by its possibility to support that feature in the future.
```
luke-biel, line 68:
```suggestion
High level service management is something that's not utilized yet, but it may be necessary to provide it in the future. The simplest example of which is to restart the service on demand by operator (administrator, for example). Such behavior is not required right now, however the choice of solution should be influenced by its possibility to support that feature in the future.
```

docs/rfc/0020_Configuration_Service_01.md
luke-biel: ```for example when service will crash and restart with different port``` k8s handles that

docs/rfc/0020_Configuration_Service_01.md
luke-biel: ```High level service management is something is not utilized yet``` well, we have some guards on k8s level that restart service on death, if that's what you mean.

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+
+#### - Security
+In context of this document, we will only discuss protocol security inside cdl network. User facing security features are ommited, but solution have to provide space for its implementation. We need to make sure that services will react only on reconfiguration requests from trusted source, and prevent man-in-the-middle injections.
```
luke-biel, line 71: User facing security features?

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+
+Various security options have to be explored for future use. Authentication and authorization are both related to the general security (and as such to the nature of configuration service), and to separate feature - "Access Groups". It requires providing "trust but verify" mechanics for both client and server. That assumes both the sender and the payload can be verified against known providers. What it entails is:
```
luke-biel, line 73: This again assumes PUSH

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+- Configuration Service has to have:
+-- CFGS's general private key
+-- operator's private key
+-- list of secrets bound to particular functionalities or access groups
+- Client of configuration service has to have:
+-- operator's private key
+-- CFGS's public key
+-- client's private key
+-- secret used to identify the permissions of the client
```
luke-biel, line 82: I'm not sure this is method of authentication we want to follow on with. Such authorization is inflexible.

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+#### - Optional Requirements:
+- providing dynamic configuration
+- providing proper authentication and authorization of configurations
+- allowing for service reconfiguration on configuration change
+- allowing for high-level service management
+- service discovery and autoprovisioning
+
+#### - Additional Assumptions
+- solution has to be neutral against deployment
+- solution has to provide api that is permissive to 3rd party services which may be written in different languages.
```
luke-biel, line 59: This, kind of, contradicts line 115: ```Configuration Service will be treated as an internal API and changes to it will not be considered breaking change.``` Not directly, but if we assume we are providing this api, we need to defend clients from breaking changes.

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+
+It is acceptable to define API during early development stage and change or expand it later.
+
+Integration with GraphQL or management panel has to be considered in later stages of the implementation to provide API for human end users.
+
+## Solutions
+All solution below will be described in detail along with its gains and losses.
+
+### Static Server
+Configuration is accessible as a URI tree served by static server. Configuration is immutable, and lives as long as the deployment (i.e. in case deployment have to change, everything is scrapped and set anew with new config). In that case, configuration can be served by any HTTP or TCP server, written in any language, including but not limited to, bash, zig, rust or an already existing tool (nginx, etcd) can be used. Any means and protocols available to serve the content are permitted for use and depends on implementor approach.
```
luke-biel, line 125: Even simplier would be k8s config mapping, as it's already present (but it'd limit us from bare metal deployments)

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+- no notifications (there are no changes either way, so those are not necessary, but still, it's a limitation)
+
+negatives:
+- inflexible
+- no security
+- no authentication
+- immutable
+- can not support reconfiguration
+- it will be increasingly harder to implement further features upon this solution
+- heavy load on the server in case of large deployments
```
luke-biel, line 146: load balancer?

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+neutral:
+- no notifications (there are no changes either way, so those are not necessary, but still, it's a limitation)
+
+negatives:
+- inflexible
+- no security
+- no authentication
+- immutable
+- can not support reconfiguration
+- it will be increasingly harder to implement further features upon this solution
```
luke-biel, line 145: why?

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+- quite a few ways to achieve this solution
+
+neutral:
+- no notifications (there are no changes either way, so those are not necessary, but still, it's a limitation)
+
+negatives:
+- inflexible
+- no security
+- no authentication
+- immutable
```
luke-biel, line 143: if server serves files, then we can easily modify these files in some cases, which allows for some mutability.

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+- simplicity
+- quite a few ways to achieve this solution
+
+neutral:
+- no notifications (there are no changes either way, so those are not necessary, but still, it's a limitation)
+
+negatives:
+- inflexible
+- no security
+- no authentication
```
luke-biel, line 142: I believe there are nginx plugins for jwt

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+- lot of work initially
+- there is a possibility that replication will have to be implemented if we want to keep multiple configuration services in sync
+
+Summary:
+- relative work: 7/10
+- maintenance work (periodic): 2/10
+- flexibility of the solution: 6/10
+- available features: 8/10
+- work needed to expand in the future: 3/10
+
```
luke-biel, line 255: I miss exploration of existing solutions.

There's zookeeper, which we already have deployed on our clusters https://zookeeper.apache.org/.

There's whole topic of cmdbs (second link sends us to some ranking):
https://www.servicenow.com/products/it-operations-management/what-is-cmdb.html
https://www.capterra.com/cmdb-software/?pricing_options=%5B8%5D&sortOrder=sponsored
https://www.dmtf.org/standards/cmdbf. But I guess we strive only for parts of the functionality.

docs/rfc/0020_Configuration_Service_01.md
```diff
@@ -0,0 +1,279 @@
...
+### Addendum
+Both p2p request-response and general rest's major problems is related to the peak load. Those issues can be alleviated in relatively similar way, both client-side and server side. Examples of which are, implementation of random sleep periods while acting on the notifications, or having multiple configuration services available. Those options will not be explored in this document because those are tightly coupled with implementation itself.
+
+While compatibility with other libp2p libraries written in different languages is stated in documentation and implied by following the same spec, it was not tested in the field by the author of this RFC.
+
+## Chosen solution
+I, personally, would opt out for libp2p solution, as the investment seems to be worth it. The amount of work required is quite high, however it can be treated as an investment, since after initial feature release is ready, almost no additional work have to be done. P2P part only have to be wrapped in common library. It is providing automatic, invisible layer to the users, and it does not have to be expanded further. All feature-related code is then common between the solution with the rest server.
```
luke-biel, line 262: One problem I see, with requirements we specified our clients have to be able to easily plug in, and with libp2p req-response it may difficult. It wasn't explored in this rfc.

docs/rfc/0020_Configuration_Service_01.md
frondeus: > What have notifications that are sent by CS and ER to do with this?
I think this paragraph does not mention notification reports sent by CS/ER/MG but notifications that "hey, configuration XYZ changed"

docs/rfc/0020_Configuration_Service_01.md
frondeus: Perhaps we want to express that "dynamism" is in our goal but can be deferred in stage 2?

docs/rfc/0020_Configuration_Service_01.md
frondeus: I think this might be connected to #636

docs/rfc/0020_Configuration_Service_01.md
frondeus: I know that libp2p is a young concept but it seems to be supported by many languages and companies: https://blog.ipfs.io/2020-06-09-libp2p-in-2020/
Is this really a problem?

docs/rfc/0020_Configuration_Service_01.md
luke-biel: Yeah, we even agreed to this, but it's not mentioned in rfc and well, it is confusing. First sentence describes ConfS as dynamic config provider.

I'd just change that.

docs/rfc/0020_Configuration_Service_01.md
luke-biel: Hmm, then it's not clear enough, we should change wording

docs/rfc/0020_Configuration_Service_01.md
luke-biel: Hmm, may be related, yeah.

docs/rfc/0020_Configuration_Service_01.md
luke-biel: It's not a problem per se - but I think it's good to note how are we gonna approach interchangeability of our services with this change - as a note for future readers. (also, probably we should decide what is the prio on making it easy to reimplement)

docs/rfc/0020_Configuration_Service_01.md
Kononnable: I think the decision on which way we should proceed should include dynamic configuration as a requirement.
It's fine if we don't implement it for now(delay implementation of 'dynamic part'), but the design and chosen technologies should allow for it to be implemented. Otherwise, we would have to do all the work from scratch.

In other words - we can delay implementation, however, design decisions should not block future implementation. If we're sure right now that dynamic configuration will be a feature later on we have to make design decisions with this feature in mind. 
As those are requirements for design, not the implementation I agree that this point should not be marked as optional.

docs/rfc/0020_Configuration_Service_01.md
Kononnable: Based on that article(I didn't do research on the current state):
Support for
- go
- js
- rust
- cpp
- py
- jvm

Not all major language families are covered here(e.g. dotnet). As for Python and JVM implementations are marked as heavy work in progress and are not recommended to be used in production.

Is it a big problem? Hard to say, probably no one will be interested in implementing theirs custom CDL components for the time being. However, this is a risk for the future to be at least acknowledged - the current state of some of those libraries is not good and there is no guarantee that they won't be just abandoned.

docs/rfc/0020_Configuration_Service_01.md
esavier: todo : clarify

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
esavier: fixed in next commit

docs/rfc/CDLF-0000F-00-rfc-01.md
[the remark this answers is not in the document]
esavier: also fixed in next commit, resolving

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: yeah, we discussed that and it appears that there  are options to secure the access for users. Therefore I am marking this as resolved :)

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: removed altogether, changed whole footnotes

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: i think i resolved this ?

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: also fixed :)

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: done?

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: fixed !

docs/rfc/0020_Configuration_Service_01.md
esavier: done in new update

docs/rfc/0020_Configuration_Service_01.md
esavier: no, there are two security features described, configuration tags, and protocol security. Access groups are mentioned elsewhere in the document (look for tags, it gives a gist of what i ment).
cfgs is so important that i want to prevent configuration from malicious services being pushed, so we need to authenticate the configuration message and confirm its source (mitigate mitm attacks for example)

docs/rfc/0020_Configuration_Service_01.md
esavier: sorry, yeah i fixed that,
I meant something else under "optional" in that case

docs/rfc/0020_Configuration_Service_01.md
esavier: this is for now a result of implementation. Currently we have to kill stuff to reconfigure it in 80% of cases, but there will be different management options later, like switching access groups etc. that will require only permission updates and not necessarily killing stuff.

docs/rfc/0020_Configuration_Service_01.md
esavier: its the api that faces user, whatever its grpc, graphql or rest, its different from service side, it requires proper user authorization etc.

docs/rfc/0020_Configuration_Service_01.md
esavier: those are components of the authorization and its only on protocol level.
how user is authenticated/authorized is different thing not in the scope of this document (we do not care right now about it tbh). We need to ensure that config we are getting is not malicious or from malicious source. Thats all. I updated the doc with examples.

docs/rfc/0020_Configuration_Service_01.md
esavier: agree, but yeah, tackled in 636,
does it resolve it?

docs/rfc/0020_Configuration_Service_01.md
esavier: as written, different tools may be used, nginx, etcd, k8s, even nfs share if someone feels adventurous :D
honestly, i wrote simple tcp server using bash and netcat back in the days and it also should work. This is a concept, not a specific solution. There are so many options that it would require yet another rfc, probably.

docs/rfc/0020_Configuration_Service_01.md
esavier: i removed it after reading what konon wrote. 
This is hard to graps but this is not concrete solution, whats common is that we have central server that will have to take care of all of the requests. I do not know if load balancer can help since its a mapper from one to many, where one is requester and many is server (chooses the most under loaded instance). In case of single central server we need to make a lot of optimizations on the request part of the services, especially that there are no notifications about changes and no fine-tuning from top to bottom. only clients can be optimized here.

docs/rfc/0020_Configuration_Service_01.md
esavier: again, depends on the solution, but for example if we will go with the simplest solution of sharing file directory, there is initially no way of preventing unauthorized access to the configuration, everyone may read everything. There is also no way of assigning specific pools to specific groups i.e. i won't be able to say, i want from pool of 10 DRs for only 3 of them to serve cold-storage and the rest to work on hot-storage etc. More advanced features usually needs more stable and advanced groundwork to already be there, so simple solutions, while providing MVP faster, will make it harder to expand upon.

docs/rfc/0020_Configuration_Service_01.md
esavier: yeah but it requires restart of for example k8s deployment. 
what you describing is written in second part, "Central Server"
also what you mentioned is reflected by large maintenance score, since there will be a lot of manual work with this

esavier: mutability with static server requires from us something like secondary service that will interact with this static server, and this is the same as central server with extra steps

docs/rfc/0020_Configuration_Service_01.md
esavier: again, no user auth, only protocol security.

docs/rfc/0020_Configuration_Service_01.md
esavier: next update will have this, i need to read some of this stuff :)

docs/rfc/0020_Configuration_Service_01.md
esavier: libp2p in rust is very stable and used by many projects,  although for neferious resons (damn those blockchains) so i wouldn't worry about it,
for the most basic functionality we just need req-response endpoint to be there and even incomplete libp2p implementations have this
worst case scenario the service will ask for config over graphql api or "user-facing" api, whatever we decide it to be

docs/rfc/0020_Configuration_Service_01.md
esavier: about zookeeper, its coordination tool, not configuration, and should not be used as one. There a lot of problems regarding zookeeper but before all of this, it's just not the purpose of this tool to provide initial configuration. For this purpose, a better choice would be simply redis instance. 
Zookeeper works literally from outside perspective and in our case also, will look like a KV database with extra steps. Usually Zookeeper is used not as configuration provider but discovery service and communication bus though. Even for Kafka its used (past tense, kafka is leaving zookeeper ecosystem) in 4 places and only one is configuration (and even in this case its partial) : leader election, broker discovery, ACL, and topic configurations. Even Kafka's configuration is not maintained inside zookeeper.

esavier: As for the CMDB, yes its a configuration service specification but for organizations, i.e. keeping track of inventory and other boring stuff. If you find the documents its literally CDL on very low-grade crack :D

frondeus (approved):

frondeus:
I Will merge it in next 2 hours if there are no objections

docs/rfc/CDLF-0000F-00-rfc-01.md
esavier: okay, its solved from my udnerstanding ?

PR #669, opened 2021-07-28
fix: Add timeout handlers for RPC

smores56:
Need someone to double-check why this isn't working for me. The timeout has been reduced, but I can't get the error message to change for some reason.


esavier:
it sounds like something i may like

esavier (dismissed):

esavier:
so after the discussion yesterday we need to make the error less leaky

esavier (approved):

PR #671, opened 2021-07-29
docs: rfc: 0017 fixes

esavier:


docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -81,55 +95,45 @@ enum LogicOperator {
...
 }
 ```
-UUID order - base_relation_parent_id, base_relation_child_id, (relation_x_child_id)*
-relation_x - relation with identifier value x
 
-Why single array(Vec) of UUIDs, not Vec<Vec<UUID>> or something similar to distinguish rows? Nested vectors means a lot of small allocations which are terrible for cpu cache. Using rust iterators allows us to handle both solutions the same way, so this optimization shouldn't be problematic from code side.
+The new response format will provide UUIDs in the following order:
+1. relation's parent's ID
+2. relation's child's ID
+3. array of relation's children's IDs
```
esavier, line 123: this is an issue not a solution.

frondeus: I agree with that, can we change it?

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -6,18 +7,22 @@ Title           : Edge registry - Resolving relation trees
...
 - TreeQuery - part of Edge registry API responsible for generating relation graphs for provided recipe and filtering criteria
+- RDBMS - relational database management system
 
 ## Motivation
-Our current process on resolving relation graphs is far from ideal - e.g. it allows querying from only one side of the relation, doesn't support complex filtering. Current implementation doesn't cover all of the features required by the materialization phase II(e.g. complex filtering).
-This document's purpose is to pass on the way behavior can be handled. Struct names and similar concepts are treated as implementation detail which may be changed during implementation.
+Currently, Edge Registry supports a list of relations that together creates non-cyclic graphs. However, there is no way to differentiate, which direction the links in the graph are pointing to (i.e. parent or a child objects) and only taking into consideration subobjects. In addition, there is no way to specify a filter composed of multiple components, bound together with usage of logical operators like `or` or `and`. Complex filters are a requirement for proper materialization support.
```
Kononnable, line 21: You're talking about tree query requests, not Edge registry. Edge registry supports all of those things(stores the data), tree query request allows fetching only some part of that data(is limited).

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -49,12 +54,21 @@ struct SchemaRelation {
 }
 ```
 
-## Proposed changes
+## Proposed formats:
```
Kononnable, line 57: Mixed style - some section headers have `:` others don't. I think it's more common to not put colons in section names(the same conversion is used in our book).

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
```diff
@@ -49,12 +54,21 @@ struct SchemaRelation {
...
+### TreeQuery request:
 
-### TreeQuery request
+```rust
+type Identifier = Option<NonZeroU8>; // id from TreeQuery->Relation, 0 for base_relation
```
Kononnable, line 61:
```suggestion
type Identifier = Option<NonZeroU8>; // id from TreeQuery->Relation, None for base_relation
```

docs/rfc/0017_Edge_registry_resolving_relation_trees_01.md
esavier: we talked with Łukasz about it, there is question about usability of this struct, but as well we can implement into() to something implementing `iter` but i do not know what's the plans around it

Kononnable: Our RFCs are (at least for me) a description of ideas. They describe the process, the format of data is an implementation detail(unless it's explicitly marked as important because of being part of end-user API). Therefore they don't limit implementation. As we don't update RFCs after they're merged and we cannot assume they reflect the current state of the feature it's fine to leave it as is and decide on a particular format during implementation. If you prefer to change it now - go for it, it doesn't change the idea.

Kononnable (approved):

frondeus (approved):

PR #684, opened 2021-08-09
fix: Revert unexpected change in DB export feature

frondeus:


crates/schema-registry/Cargo.toml
```diff
@@ -45,3 +45,4 @@ tracing-futures         = "0.2.5"
 
 [dev-dependencies]
 maplit = "1.0.2"
+diff_assert = { git = "https://github.com/frondeus/diff-assert.git" }
```
luke-biel, line 48: Is there a reason why we use fork that is 5 commits behind the master?

frondeus: That's a good catch! To be honest I was 99% sure that I made a change there and it is not in the upstream repo. I'm gonna change it to the original repository.

luke-biel (approved):

esavier:
what integration tests are you planning?

PR #688, opened 2021-08-11
fix: rfc: Minor context fix

esavier:
smoll fix :3


frondeus (approved):

Kononnable:
Not sure why we need that/should we do this.
RFC are up-to-date on the day of merge, after that we decided not to change them.

esavier:
i just removed what was added in following PR's
task was done and rfc was done

PR #699, opened 2021-08-16
docs: materialization types - rfc

luke-biel:
[rendered](https://github.com/epiphany-platform/CommonDataLayer/blob/49967be40ee5e7b463d8c9b8508b2aeb3f034163/docs/rfc/0020_Materialization_Types.md)

### DIFF
--- a/docs/rfc/0020_Materialization_Types.md
+++ b/docs/rfc/0020_Materialization_Types.md
@@ -1,0 +1,314 @@
+# Front Matter
+
+```
+Title           : Materialization field types support
+Category        : Feature
+Author(s)       : Łukasz Biel
+Team            : CommonDataLayer
+Created         : 2021-12-08
+Deadline        : 2021-12-08
+CDL feature ID  : CDLF-00018-00
+```
+
+Related:
+
+* https://github.com/epiphany-platform/CommonDataLayer/issues/656
+
+# Glossary
+
+* TS - timeseries
+* QSTS - `query-service` for timeseries
+* QS - `query-service` for documents
+* ES - elasticsearch
+* PSQL - PostgreSQL
+* SR - Schema Registry
+* Blob - `any` data, eg. pdf report, ms office spreedsheet, jpeg image, that CDL shouldn't preprocess in any way
+* ADT - algebraic data type ([wiki][wiki-adt])
+* ConfS - configuration service, central repository of CDL configuration, currently in rfc phase.
+
+# Summary
+
+TODO
+
+# Configuration layer
+
+At the moment of writing this rfc, data is stored in cdl in the following format:
+
+#### Documents
+
+JSON objects. They are not indexed by any field and querying for only parts (fields or sub-documents)
+requires fetching whole JSON and then performing operations on it.
+
+#### Timeseries
+
+Timeseries data is managed by databases, often requires special queries on QSTS side, but most likely can query for
+specific fields.
+
+#### Blob
+
+Blobs are stored as is. You cannot perform operations on it.
+
+> SIDENOTE:  
+> Other storage types are included for future reference. Currently, materialization is limited to Document Store.  
+> If that was to change, new document would have to be created, exploring the integration.  
+> From this point on we are gonna focus on document storage only.
+
+JSON has fairly limited type system. We can read more about it in [JSON Schema spec][json-schema-basic-types]. JSON
+schema is relevant there due to it being used in SR.
+> SIDENOTE:  
+> While it's used in SR, it's not really "used". It is planned to be incorporated in validation.
+> [rfc 0019][rfc-0019] talks about simpler format, as a replacement and for the sake of Materialization I will explore expanding it later.
+
+| type name | description |
+|---|---|
+| number | any number of any precision |
+| string | Unicode string of characters of any length |
+| boolean | "true" or "false" keyword |
+| null | a null value, or, to be precise, lack of thereof |
+| object | a json key-value, unordered, map |
+| array | an ordered collection of records, can hold mixed types |
+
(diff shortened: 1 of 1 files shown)


docs/rfc/0020_Materialization_Types.md
```diff
@@ -0,0 +1,449 @@
...
+## on-demand materializer
+
+On demand materializer serves data over grpc, returning materialized rows mapping field name to stringified json at the moment.
+
+We should update mentioned format to return types that resemble view type definition in either a nested grpc message or map where keys are `.` delimited paths.
+> SIDENOTE  
+> '.' delimited paths are borrowed from ES, where this is how columns in internal indexes are stored.
+
+### nested messages
+We should map view types to grpc types via mapping:
```
Kononnable, line 377: Not sure we can cover that on grpc level - proto files are created when there is no information of view column types. That means that we would have to create some enum, write code to convert values, etc. The same work will have to be done by the client, it might be easier to pass results as JSON - clients know view definition so they can convert the values if they need to.

docs/rfc/0020_Materialization_Types.md
```diff
@@ -0,0 +1,449 @@
...
+bool -> bool
+array[basic_type] -> `repeated`
+object -> `message` # recursive
+
+nullability: `optional`
+```
+This will require us to use `oneof` operator and will result in messages sent from one service to another being nested.
+
+### strings
+We will send all data as string over grpc and fields will be in `path.to.deep.field` format (to ensure everything is flat), eg.:
```
Kononnable, line 391: Why does this have to be flat? Can We reuse the same approach as with the Postgres materializer(at least for now)?

docs/rfc/0020_Materialization_Types.md
```diff
@@ -0,0 +1,449 @@
...
+> SIDENOTE:  
+> We can skip "type" field and require users to use views to deduce field types.
+
+### Opinion
+We should aim for first option if possible, as it looks cleaner and provides better network performance.
+
+# Other options
+
+## We can assume everything is nullable
+As in title, CDL may assume that nullability is by default, thus we have less work in case of PSQL materialization. We still have to be able to parse
```
Kononnable, line 423: Unfortunately, nullability is not a simplification in many cases. 
In the case of nullable repository data - we have to handle nullability for all of the defined operators(computed properties), decide if they propagate etc. Another problem is filtering similiar to SQL `where column in (null)` - this will behave differently depending on sql engine settings.
As for nullability in materialization output - the same problems will exist, but responsibility for handling them will be passed to the users.

docs/rfc/0020_Materialization_Types.md
```diff
@@ -0,0 +1,449 @@
...
+ES initially treats objects differently to PSQL.
+
+We can later expand this type list with per database custom types (and type casting/mapping) but this requires schema language extension or parsing strings in json object.
+
+SR emits a notification when view is added, initiating materialization for it and provisioning the database.
+
+GRPC between OB and Materialization is:
+```
+i64 -> int64
+f64 -> double
```
Kononnable, line 94: Do we care about types with arbitrary precision e.g. decimal(5,2)? 
What is proposed allows only saving integers, without precision value or use 'non-exact' types. In other words, users have to handle precision values in their code(e.g. multiply data by powers of ten) or use type without arbitrary precision which can store the wrong value instead of throwing an error.

I'm fine with not handling that now but I guess that should be stated publicly.

docs/rfc/0020_Materialization_Types.md
luke-biel: For now - yeah. For later we are missing type info on nested fields. Depends on how important it'll be

docs/rfc/0020_Materialization_Types.md
luke-biel: This solution proposes that enum. We can either use enum or send strings and I prefer enum over string.

docs/rfc/0020_Materialization_Types.md
luke-biel: Decimals and other types can be handled later. This only proposes base types. (unless some client states otherwise)

esavier (dismissed):

docs/rfc/0020_Materialization_Types.md
```diff
@@ -0,0 +1,449 @@
+# Front Matter
```
luke-biel, line 1: Newline after title only if next is another title

Titles should have title case
	
Titles don't have colons

https://raw.githubusercontent.com/epiphany-platform/CommonDataLayer/57156a54a198a762978a4cf1d739b1c5ebb20920/docs/rfc/0000_Template_00.md

docs/rfc/0020_Materialization_Types.md
luke-biel: We could also send whole view as one string instead of map<string, nested>, thinking of it now.

docs/rfc/0021_Materialization_Types.md
```diff
@@ -0,0 +1,436 @@
...
+Objects, or complex arrays, are much more problematic, and we have to review whether we need them. I'll get to this in a moment.
+
+PSQL mapping in this case:
+```
+`String` -> `psql::text`
+`bool` -> `psql::bool`
+`i64` -> `psql::int8`
+`f64` -> `psql::float8`
+`null` -> COLUMN IS NULLABLE
+`array` -> type[]
```
frondeus, line 244: Do we still want to cover array types here?

docs/rfc/0020_Materialization_Types.md
luke-biel: We decided to not support nullability for the time being

frondeus (approved):

PR #701, opened 2021-08-17
fix: lock versions of postgres and vm in compose

luke-biel:
We have to lock version of vm to 1.63 as 1.64 introduces some changes that mess up our tests


Kononnable (approved):
Not sure if that should be `fix` or `chore` as it fixes CI, not the end product.

PR #702, opened 2021-08-17
fix: human friendly message in API

### LINKED ISSUE #522: bug: API returns human-friendly error message
**Describe the bug**
API gets stuck on view insertion causing timeout. Message included in the error response is too detailed and technical.

**Expected behavior**
Error message should be human readable and generic. It should not include any technical details.
Instead, user can always look into logs to see details.

Acceptance criteria:
* [ ] Error returned in graphQL response should not contain any technical details
* [ ] Response needs some kind of span/id information to pinpoint where in logs we should look for an error (breadcrumbs)

frondeus:
fixes #522


esavier (approved):

esavier:
merging today at the end of the day if there are no other issues

crates/api/src/schema/mutation.rs
```diff
@@ -71,7 +71,8 @@ impl MutationRoot {
                     .collect(),
             })
             .await
-            .map_err(|source| rpc::error::ClientError::QueryError { source })?
+            .map_err(Error::query)
```
smores56, line 74: I know these are cleaner, but wouldn't it just be simpler to have these in a single map?

smores56 (approved):

PR #704, opened 2021-08-19
feat: materialization to elasticsearch

luke-biel:


esavier (approved):
cooooool byt we need to add deployment for that xD
you went kinda too fast with it, first we need to add docker-compose stuff, i also need to add it to the playground later

esavier (commented):
lets wait for deployment additions

crates/materializer-general/src/lib.rs
```diff
@@ -166,3 +167,21 @@ impl MaterializerImpl {
         Ok(())
     }
 }
+
+pub async fn materializer_plugin(
```
smores56, line 171: Shouldn't this go in the settings_utils?

crates/materializer-general/src/plugins/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn validate_options(&self, options: Value) -> anyhow::Result<()>;
+    async fn upsert_view(
+        &self,
+        view: MaterializedView,
+        view_definition: FullView,
+    ) -> anyhow::Result<()>;
+}
+
+#[derive(Debug, Serialize)]
+struct RowDefinition {
```
smores56, line 26: This doesn't seem like an appropriate place for this struct definition. Is there somewhere else it can go?

crates/dto/src/materialization.rs
```diff
@@ -447,6 +447,11 @@ pub struct PostgresMaterializerOptions {
     pub table: String,
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct ElasticsearchMaterializerOptions {
+    pub index_name: String,
+}
+
```
smores56, line 454: These aren't as much DTO's as they are config structs. Consider moving them out of here.

smores56 (dismissed):

luke-biel:
> cooooool byt we need to add deployment for that xD
> you went kinda too fast with it, first we need to add docker-compose stuff, i also need to add it to the playground later

I have D-C ready, couldn't open PR today though

crates/materializer-general/src/lib.rs
luke-biel: Can't. Types returned are crate specific :/ All I could do is let this accept array of constructors, but that seems like an overkill.

crates/dto/src/materialization.rs
frondeus: This is DTO, it is passed via ViewDefinition, not via Setting. Each view can use a different index name.
The same MaterializerOptions trick we use for Postgres, where you can define to what table view should be materialized.

luke-biel:
https://github.com/epiphany-platform/CommonDataLayer-deployment/pull/11 - adding ES to d-c

esavier:
docker compose is in the deployment, i am okay with whatever <3

crates/materializer-general/src/lib.rs
esavier: okay, we talked about it a bit and it seems like kinda overkill for now.
we will do it when we get it lets say :)

crates/dto/src/materialization.rs
esavier: okay, also resolving this, we debated it for a short while and we agree this kinda can stay here for now

esavier (approved):

PR #729, opened 2021-08-30
chore: workflow: skip on doc commits

### LINKED ISSUE #712: bug: stop doing unit tests on document commits
AC:
  - add scripting or something?

esavier:
fixes #712


.github/workflows/test-cdl-config.yml
```diff
@@ -3,11 +3,35 @@ name: Configuration
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 32
+
+      - id: decision
+        run: |
+          echo "commit file list:"
+          if git --no-pager diff --name-only HEAD~1 | grep -v .sh | grep -v .md
```
frondeus, line 21: Are you sure we want to ignore "sh"?
What if we change something in build.sh which will cause a production error that we won't detect because our tests will be ignored?

esavier: i think i do,
this is the full list of shells we have:

```
./.github/scripts/install-executable-from-git.sh
./.github/scripts/release.sh
./build.sh

./examples/python-cdl/cdl.sh
./deployment/helm/helpers/minikube-setup-jaeger.sh
./tests/run-tests.sh
./tests/prepare-test-env.sh
./e2e/test.sh
./CDL.Tests/build.sh
./CDL.Tests/test.sh
./crates/object-builder/tests/review.sh
```
and only the first 3 are used actively in the workflows, although are updated once every lifetime. 
first 2 are not tested at all, the 3rd one is tested passively by usage,

frondeus (approved):

.github/workflows/test-cdl-config.yml
Kononnable: `./tests/run-tests.sh` - used by test-component.yml
`./e2e/test.sh` - used by test-deployment.yml
`./CDL.Tests/build.sh` - used by test-deployment.yml
`./CDL.Tests/test.sh` - used by test-deployment.yml

I think it's better to run sometime a test pipeline that is not required than to skip one when it should run. Anyway, It's not likely that those files will be changed often on their own, without changing other files that will trigger the workflow.

.github/workflows/test-cdl-config.yml
```diff
@@ -3,11 +3,35 @@ name: Configuration
 on:
   pull_request:
```
luke-biel, line 4:
```suggestion
on:
  pull_request:
    paths-ignore:
      - '**/*.md'
```

luke-biel: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths for reference

PR #662, opened 2021-07-26
chore: fix CI

Kononnable:

PR #731, opened 2021-08-30
chore: test: update k8s deployment

esavier:


frondeus (approved):

Kononnable:
Please run deployment tests on CI before merging. As I recall `v1.21.3` didn't work with the current setup(#662).

Kononnable (changes requested):
Blocking so no one merges that by mistake before running changed workflow to test if it works.

Kononnable (approved):

PR #739, opened 2021-09-01
feat: Simplify schema definitions

frondeus:


crates/utils/types/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
frondeus, line 1: Do we need `utils/types`? Cant we reuse `cdl_dto` types? Or perhaps we need to rename `cdl_dto` into `cdl_types`.

esavier: yeah the second one, i talked with sam yesterday about it and DTO means data transfer objects, sam added just shared types that should be "shared" but are not "transfer"
so
idea is to just move it to utils crate, but it takes some time and for now, adding it as this is okay imo

frondeus: Okay but then at least lets not categorize it as "utils" and move it to main `crates/` folder.
Id say it contains some kind of business/domain knowledge therefore its not agnostic "utils"

crates/schema-registry/src/db.rs
```diff
@@ -241,21 +240,22 @@ impl SchemaRegistryDb {
...
+                id, name, schema_type, insert_destination,
+                query_address, definition) \
              VALUES($1, $2, $3, $4, $5, $6)",
             &new_id,
             &schema.name,
             &schema.schema_type as &rpc::schema_registry::types::SchemaType,
             &schema.insert_destination,
             &schema.query_address,
-            &schema.definition,
+            &schema.definition as _,
```
Kononnable, line 258: Some type conversion problems, or why `as _` in some many places(including sql)

crates/schema-registry/src/validation.rs
```diff
@@ -0,0 +1,233 @@
...
+                            return Err(RegistryError::InvalidData(format!(
+                                "Value was null or missing: Path `{}`",
+                                field_path.join(", ")
+                            )));
+                        }
+                    }
+                }
+            }
+        }
+        SchemaFieldType::Array(item_type) => {
```
Kononnable, line 88: Same as above.

crates/utils/types/src/schemas.rs
```diff
@@ -0,0 +1,102 @@
...
+    Array(Box<SchemaFieldDefinition>),
+}
+
+impl TryFrom<SchemaFieldDefinitionRpc> for SchemaFieldDefinition {
+    type Error = anyhow::Error;
+
+    fn try_from(definition: SchemaFieldDefinitionRpc) -> Result<Self, Self::Error> {
+        let old_type = definition.field_type;
+        let mapped_type = match old_type.field_type {
+            0 => SchemaFieldType::Scalar(
```
Kononnable, line 35: Very ugly :| 
What if someone changes the enum definition(proto)? Take a look at `rpc_enum` it looks like it's solving the issue.

crates/schema-registry/src/db.rs
smores56: This is unfortunately required in some cases with SQLx: https://github.com/launchbadge/sqlx/blob/2eb4ff8713f24c5fc7f7fb6ad1921a87d0048cf6/examples/postgres/json/src/main.rs#L80

crates/api/Cargo.toml
```diff
@@ -18,6 +18,7 @@ path = "src/lib.rs"
 cdl_dto     = { path = "../dto" }
 misc_utils  = { path = "../utils/misc" }
 rpc         = { path = "../rpc" }
+types       = { path = "../utils/types" }
```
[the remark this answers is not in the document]
smores56: Moved to a normal crate location.

crates/api/src/types/schema.rs
```diff
@@ -22,7 +24,7 @@ pub struct FullSchema {
     /// Whether this schema represents documents or timeseries data.
     pub schema_type: SchemaType,
     /// The format of data stored under this schema.
-    pub definition: Value,
+    pub definition: Json<HashMap<String, SchemaFieldDefinition>>,
```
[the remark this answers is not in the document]
smores56: Extracted to a common type alias.

crates/rpc/proto/schema_registry.proto
```diff
@@ -213,4 +219,27 @@ message SchemaType {
     required Type schema_type = 1;
 }
 
+message ScalarType {
+    enum Type {
+        Bool = 0;
+        String = 1;
+        Integer = 2;
+        Decimal = 3;
```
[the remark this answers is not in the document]
smores56: Updated!

crates/utils/types/src/schemas.rs
smores56: I messed with `rpc_enum` the first time a good deal, and wasn't successful after a few hours of mucking around in getting it to work correctly between a bunch of different type/compilation errors, for some reason. I'll give it another go.

.gitignore
```diff
@@ -7,4 +7,5 @@ __pycache__
 tests/rpc/
 docs/mdbook-plantuml-img/
 .tmp
+**/.env
```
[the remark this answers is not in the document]
smores56: I thought it was minor enough to be insignificant, but I removed it from this PR.

crates/utils/types/src/schemas.rs
smores56: Fixed, by creating a new enum message wrapper type. I was avoiding it because it seemed like a hack-y solution, but it works well enough.

crates/schema-registry/src/validation.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    Ok(())
+}
+
+fn validate_data_inner(
+    data: &Value,
+    type_: &SchemaFieldDefinition,
+    path: &[String],
+) -> RegistryResult<()> {
+    if data.is_null() {
```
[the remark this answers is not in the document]
smores56: I've been thinking about this for a bit, and it seems like the move (IMO) is to keep the option for nullability. Missing values are fundamental to representing data, and it seems like a very strong limitation to place on data types. I know we are not really using Schema Types for validation, but in light of the halt of active development, this seems like a bad feature to leave partially finished. Also, it should be simple to treat any nullable value as arbitrary JSON from the perspective of materialization, if we really can't afford to introduce nullability. What are your thoughts?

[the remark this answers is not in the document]
Kononnable: Sorry for the delay. As I recall no-nullability was meant just as a temporary state - MVP. As CDL development is getting halted I'm not sure if this does matter. Once(if) someone gets back to it they will have to decide what to do with the situation either way. For me, it's fine to leave values as non-nullable as well as to allow them to be nullable on schema level but act on them like any JSON column(no transformation operators) in materialization(if it's properly documented when this part of materialization is done).

PR #742, opened 2021-09-02
fix: Marry 0019 rfc with 0022

frondeus:
Please double-check the whole 0022 if its integrity is not compromised.

### DIFF
--- a/docs/rfc/0019_Simplify_Schema_Definitions_01.md
+++ b/docs/rfc/0019_Simplify_Schema_Definitions_01.md
@@ -32,7 +32,7 @@ Schema:
 
 ScalarType:
 ```
-Bool | String | Integer | Decimal | Any // Any represents arbitrary JSON data
+Bool | String | Integer | Float | Any // Any represents arbitrary JSON data
 ```
 
 ObjectFieldDefinition:
@@ -99,7 +99,7 @@ further on in this proposal.
 
 In an effort to maintain a similar structure to JSON data and not added unnecessary complexity, SD's will be
 represented as sum types. Beyond a variant for each of the scalars available in JSON (`Boolean`, `String`,
-`Integer`, and `Decimal`), the other two types of JSON data are `Array`s and `Object`s. `Array`s will refer
+`Integer`, and `Float`), the other two types of JSON data are `Array`s and `Object`s. `Array`s will refer
 to a single child definition, whether that be a scalar type or a different schema's definition. `Object`s will
 refer to a list of string-named fields, where the definition of each field is either a scalar or another Schema's
 definition. `Array` or `Object` child definitions can all be optional, which allows items to either be `Null`
@@ -131,6 +131,12 @@ Proposed is the suggestion to store for each database materialized data is store
 would be in the config service (really, wherever the materialization configuration data is stored).
 
 ## Further Considerations
+
+### Decimal vs Float
+Though `Float` means that it can be inprecise, there are many fields and industries where it is required. 
+CDL has to support it and can not replace it with `Decimal` (fixed-point fractional arithmetic).
+
+`Decimal` type is not required for the MVP, but can be added later as an extension to the type system.
 
 ### Impact on Other Teams/Clients
 Though validation is not currently being done on incoming data and therefore schema definitions don't matter
--- a/docs/rfc/0022_Materialization_Types_01.md
+++ b/docs/rfc/0022_Materialization_Types_01.md
@@ -32,21 +32,24 @@ This rfc aims to provide and standardize types in CDL schemas, views and materia
 # TLDR
 Schemas support
 ```
-number
+bool
 string
-boolean
+integer
+float
 object
+any (arbitrary JSON data)
 ```
 
-Any other type used within JSON schema should cause an error when view materialization is requested.
+`array` type would be treated as `any` in materialization.
 
 Views support (names are arbitrary, implementor may choose suitable naming)
 ```
-f64
-i64
-string
 boolean
+string
+i64
+f64
 object - Field::Object
+any
 ```
 
 PSQL mapping is
@@ -56,6 +59,7 @@ i64 -> int8
 string -> text
 boolean -> bool
 object -> each value resides in a column with `_` delimited json field path as a name
+any -> JSON column
 ```
 
 ES mapping is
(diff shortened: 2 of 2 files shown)


docs/rfc/0019_Simplify_Schema_Definitions_01.md
```diff
@@ -132,6 +132,12 @@ would be in the config service (really, wherever the materialization configurati
 
 ## Further Considerations
 
+### Decimal vs Float
+Though `Float` means that it can be inprecise, there are many fields and industries where it is required. 
+CDL has to support it and can not replace it with `Decimal` (fixed-point fractional arithmetic).
+
+`Decimal` type is not required for the MVP, but can be added later as an extension to the type system.
+
```
Kononnable, line 140: Not sure if we have to compare those two if we're not including decimals. It's not one or the other(and why this one), it's just one for now because of MVP rules.

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -32,21 +32,24 @@ This rfc aims to provide and standardize types in CDL schemas, views and materia
 # TLDR
 Schemas support
 ```
-number
+bool
```
Kononnable, line 35:
```suggestion
boolean
```
Or change other booleans to bool.

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -32,21 +32,24 @@ This rfc aims to provide and standardize types in CDL schemas, views and materia
...
 string
-boolean
+integer
+float
 object
+any (arbitrary JSON data)
 ```
 
-Any other type used within JSON schema should cause an error when view materialization is requested.
+`array` type would be treated as `any` in materialization.
```
Kononnable, line 43: Shouldn't we put it as schema-supported type then and mark it in mappings?

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -97,24 +101,24 @@ Blobs are stored as is. You cannot perform any operations on it (except for fetc
...
-|:----------|:-------------------------------------------------------|
-| number    | any number of any precision                            |
-| string    | Unicode string of characters of any length             |
-| boolean   | "true" or "false" keyword                              |
-| null      | a null value, or, to be precise, lack of thereof       |
-| object    | a json key-value, unordered, map                       |
-| array     | an ordered collection of records, can hold mixed types |
+| type name | description                                                             |
+|:----------|:------------------------------------------------------------------------|
+| bool      | "true" or "false" keyword                                               |
```
Kononnable, line 112: This is about JSON types, not schema types. Changing header may be required.

Also bool vs boolean again.

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -220,6 +224,12 @@ ES is able to deduce types based on insertions (first insertion to a given index
 Another main difference is that ES allows for storage of documents. While currently our PSQL implementation is flat (or to be more specific, it flattens everything to one table),
 ES can store deeply nested objects. It also keeps an index on every field in a document.
 
+### Array Type
+In ES there is no dedicated `array` data type. Any field can contain zero or more values by default, however, all values in the array must be of the same data type. In case of Schema definition type it is required and validated by `Array::item_type`
```
Kononnable, line 228: Not sure what that brings to the table if we decided not to utilize arrays for now.

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -311,25 +310,25 @@ Having types map from schema to view, we can add validation to the process of cr
 
 ## Type Storage
 > DECISION:
-> TLDR section refers to currently supported subset of json schema. In basics, we can store number, boolean, string and object (composed of 3 previously mentioned base types).
+> TLDR section refers to currently WIP Simple Schema Definition.
```
Kononnable, line 313: Change to link?

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -311,25 +310,25 @@ Having types map from schema to view, we can add validation to the process of cr
...
 > Other types in json schema are supported by CDL but not materialization.
 
 There are several options where types should be stored. In my opinion, it's view responsibility to store types of columns, and it's schema responsibility to store type of data.
-We keep current schema format, with JSON compatible types (`string`, `number`, `boolean`, `one_of null`, `array`, `object`) and require users to specify type mapping for view.
+We keep current schema format, with JSON compatible types (`string`, `integer`, `float`, `boolean`, `array`, `object`) and require users to specify type mapping for view.
```
Kononnable, line 317: JSON compatible types vs integer and float.

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -311,25 +310,25 @@ Having types map from schema to view, we can add validation to the process of cr
...
   bool <- boolean
 
 complex_type:
-  array[basic_type] <- array # will panic when array contains mixed types
+  array[basic_type] <- array # In first draft we can convert it to `any`
   object(nested) <- object
   
 nullability:
-  additional field `nullable` that maps to `one_of: any | null` # `nullable = false` will have behaviour similar to .unwrap()
+  additional field `optional` for object field definition and array
```
Kononnable, line 331: Not sure how we should handle nullability/optional in this document. Later on, we say(in the decision section) that nullability won't be implemented for now. It's just weird to read updates that mention nullability often.

Kononnable (changes requested):

docs/rfc/0022_Materialization_Types_01.md
```diff
@@ -311,25 +310,25 @@ Having types map from schema to view, we can add validation to the process of cr
 
 ## Type Storage
 > DECISION:
-> TLDR section refers to currently supported subset of json schema. In basics, we can store number, boolean, string and object (composed of 3 previously mentioned base types).
+> TLDR section refers to currently WIP Simple Schema Definition.
 > Other types in json schema are supported by CDL but not materialization.
```
Kononnable, line 314: We should no longer mention json schema if this document is based on the new approach.

docs/rfc/0022_Materialization_Types_01.md
Kononnable: Do we need to say anything about the JSON compatibility of schema here at all? It's about views after all.

